PR #132, opened 2023-09-19
String resources

JohnOberhauser:
- converting all hardcoded strings into string resources
- made some modifications to `StringFactory`
  - Adding a subclass for plurals, and one for a collection of factories

### DIFF
--- a/core/common/src/main/java/org/mozilla/social/common/utils/StringFactory.kt
+++ b/core/common/src/main/java/org/mozilla/social/common/utils/StringFactory.kt
@@ -8,7 +8,7 @@ import androidx.annotation.StringRes
  * String wrapper class that supports literals and string resources
  */
 sealed interface StringFactory {
-    private data class LiteralString(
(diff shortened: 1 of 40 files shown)


core/common/src/main/java/org/mozilla/social/common/utils/StringFactory.kt
```diff
@@ -1,35 +1,88 @@
...
 import androidx.annotation.StringRes
-import androidx.compose.runtime.Composable
-import androidx.compose.ui.platform.LocalContext
 
 /**
  * String wrapper class that supports literals and string resources
  */
 sealed interface StringFactory {
-    private data class Literal(val literalValue: String) : StringFactory
+    private data class LiteralString(
```
devotaaabel, line 11: I don't think `String` needs to be in both the factory name and the implementing class names. It feels kinda redundant/verbose

timc-mozilla (approved):

PR #149, opened 2023-09-26
[no bug] Move Glean and Jetbrains-Python-Envs back to buildscript

travis79:
The only way I ended up getting this to work required moving the import back to the buildscript and applying the plugins directly. That step also required relaxing the repositories mode from failing on project repository definitions to "prefer settings". 

I'm not particularly happy with how this integrates with the newer "DSL" gradle syntax and will be filing a bug to investigate improvements to Glean docs and the gradle-plugin to help with this.

### DIFF
--- a/core/analytics/build.gradle.kts
+++ b/core/analytics/build.gradle.kts
@@ -1,9 +1,17 @@
+buildscript {
+    dependencies {
+        classpath("gradle.plugin.com.jetbrains.python:gradle-python-envs:0.0.31")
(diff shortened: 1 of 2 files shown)


devotaaabel:
Thanks for your help @travis79. AFAIK, @timc-mozilla has this fixed on his branch, so I'm gonna close this. This is just a hunch that @JohnOberhauser had brought up, but I think there's limitations with the new DSL that are possibly caused by the plugin not being listed in the [gradle plugin portal](https://plugins.gradle.org/).

PR #151, opened 2023-09-26
Profile overflow button with mute, block and report

JohnOberhauser:
- Adding an overflow menu button to the profile screen for other user's profiles
- Overflow options:
  - mute
  - unmute
  - block
  - unblock
  - report


core/domain/src/main/java/org/mozilla/social/core/domain/GetDetailedAccount.kt
```diff
@@ -0,0 +1,72 @@
...
+    ): Flow<Resource<T>> = flow {
+        emit(Resource.Loading())
+        val mutex = Mutex(true)
+        var exception: Exception? = null
+        coroutineScope.launchSupervisor {
+            try {
+                val accountJob = async { accountRepository.getAccount(accountId) }
+                val relationshipJob =
+                    async { accountRepository.getAccountRelationships(listOf(accountId)) }
+                val account = accountJob.await()
```
devotaaabel, line 44: These happen in parallel if you use `awaitAll` instead

core/common/src/main/java/org/mozilla/social/common/Resource.kt
```diff
@@ -0,0 +1,14 @@
+package org.mozilla.social.common
+
+sealed class Resource<T> {
```
devotaaabel, line 3: Can you plz add a quick comment on this with a description of the expected usage, if the idea is for it to be used generally?

devotaaabel (approved):

core/domain/src/main/java/org/mozilla/social/core/domain/GetDetailedAccount.kt
JohnOberhauser: They happen in parallel already with the `async` blocks.  `async` blocks start when created, not when `await` is called.  I'm not sure `awaitAll` has any advantage over `await` here.  Each job returns different types, so it might make things more complicated actually.

devotaaabel: Yup good call, I didn't notice the `async` lol. There's the benefit of both jobs being cancelled if one fails using `asyncAwait`, but agreed that it's not necessary.

PR #156, opened 2023-09-28
follow and unfollow button

JohnOberhauser:
- Adding a follow and unfollow button the the profile screen
- Adding an edit button when viewing your own profile (not functional yet)
- Moving some stuff around in the profile screen to make room for the new buttons
- Created a `MoSoButton`

### DIFF
--- a/feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
+++ b/feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
@@ -403,9 +403,10 @@ private fun Header(
     accountUiState: AccountUiState,
     accountInteractions: AccountInteractions,
 ) {
-    HeaderLayout(
-        modifier = modifier.fillMaxWidth(),
-        headerImage = {
+    Box(
+        modifier = modifier.fillMaxWidth()
+    ) {
+        Column {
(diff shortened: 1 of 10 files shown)


feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
```diff
@@ -405,77 +397,106 @@ private fun QuickFunctions(
...
-        ) {
-            AsyncImage(
-                model = headerUrl,
-                contentDescription = null,
-                contentScale = ContentScale.Crop,
-                modifier = Modifier
-                    .fillMaxWidth()
-                    .height(120.dp)
-            )
+    Layout(
```
devotaaabel, line 468: I'd like to hear more about why this was used instead of a combination of rows/columns. I'm worried that this is rigid, and will only work with the inputs in a specific order

JohnOberhauser: Spoke about this in slack.  Removed the layout

JohnOberhauser:
Merging with the assumption that the only change requested was the layout thing.  Let me know if there is anything else and I can fix it.

PR #161, opened 2023-09-29
Feature glean

timc-mozilla:
adding glean to app with test metric reporting


settings.gradle.kts
```diff
@@ -2,16 +2,43 @@ pluginManagement {
     includeBuild("build-logic")
     repositories {
```
JohnOberhauser, line 3: I think this file needs to be cleaned up

devotaaabel (approved):

settings.gradle.kts
timc-mozilla: can you explain what you mean by this please?

JohnOberhauser: Sorry for the vague comment.  It was hard to know what could be removed without touching the code myself.  I added a commit that cleans this up.  It deletes unnecessary changes, moves some stuff to the analytics gradle, and adds dependencies in the libs.versions file

JohnOberhauser (approved):

PR #162, opened 2023-09-29
Share account

JohnOberhauser:
- Adding a share option to the account overflow menu
- Creating a MoSoDropdownMenu
  - Replacing DropdownMenus throughout the app with the Moso one
- **Using a custom top bar for the account screen that has the overflow menu**
  - Possibly a controversial choice.  Let me know what you think.
  - Did this so that top bar logic specific to this screen exists in the account module.
  - Another option would be to move the overflow button somewhere else when viewing the profile tab
Screenshot_20230929_170623


feature/account/src/main/java/org/mozilla/social/feature/account/AccountNavigation.kt
```diff
@@ -37,7 +37,7 @@ fun NavGraphBuilder.accountScreen(
     composable(
         route = MY_ACCOUNT_ROUTE,
     ) {
-        AccountScreen(
+        AccountRoute(
```
devotaaabel, line 40: I know we've been more or less following this pattern since the beginning, but I'm a little skeptical about it... 

1. I don't think the navigation logic should be in the module- that's implementation logic specific to our implementation (and won't make sense to anyone else that imports this module). Especially if we're going to be exporting modules eventually
2. I also find the `<name>Route` naming convention kinda confusing, and would rather just have the `composable(` text in the navigation code, because it's self-describing in itself, and I feel like we're kind of obscuring useful information by encapsulating it in a function. 

Curious other people's thoughts

feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
```diff
@@ -174,69 +178,90 @@ internal fun AccountScreen(
 @Composable
 private fun OverflowMenu(
     account: AccountUiState,
+    isUsersProfile: Boolean,
```
devotaaabel, line 181: This param name is slightly ambiguous... Would `isOwnProfile` make sense?

feature/account/src/main/java/org/mozilla/social/feature/account/AccountNavigation.kt
JohnOberhauser: I'm going to add this to the TIP agenda

feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
JohnOberhauser: Works for me.  Fixed

PR #169, opened 2023-10-02
Profile loading / error states

JohnOberhauser:
- Adding a loading and error state for load account info in the profile screen
- Modifying loading / error states in the `PostCardList` so they work in more situations (especially on the profile screen)

### DIFF
--- a/core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCardList.kt
+++ b/core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCardList.kt
@@ -1,22 +1,15 @@
 package org.mozilla.social.core.ui.postcard
 
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.foundation.lazy.LazyColumn
-import androidx.compose.material3.Button
 import androidx.compose.material3.CircularProgressIndicator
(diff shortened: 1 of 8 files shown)


core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCardList.kt
```diff
@@ -26,40 +21,37 @@ import androidx.paging.compose.itemKey
...
+ * @param headerContent content that will show above the list
+ */
 @Composable
 fun PostCardList(
     feed: Flow<PagingData<PostCardUiState>>,
     errorToastMessage: SharedFlow<StringFactory>,
-    reccs: List<Recommendation>? = null,
     postCardInteractions: PostCardInteractions,
     enablePullToRefresh: Boolean = false,
+    fullScreenRefreshStates: Boolean = false,
```
devotaaabel, line 49: Can you update the tense of the param names on 48-49 plz? Something along the lines of `pullToRefreshEnabled` and `isFullScreenRefresh`

feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
```diff
@@ -78,35 +81,25 @@ internal fun AccountRoute(
...
     MoSoToast(toastMessage = viewModel.errorToastMessage)
 }
 
 @Composable
-internal fun AccountScreen(
-    account: AccountUiState,
-    showTopBar: Boolean,
+private fun AccountScreen(
+    resource: Resource<AccountUiState>,
+    showCloseButton: Boolean,
```
devotaaabel, line 102: Similar to other comment- for consistency with unidirectional data flow, I think it's helpful for params to be phrased as if they are describing the UI state as attributes, instead of instructing the UI what to do. So here, can you rename this something like `closeButtonVisibility` or `isCloseButtonVisible` plz?

devotaaabel (approved):
LGTM, just some NITs about naming conventions.

PR #172, opened 2023-10-04
Code cleanup

JohnOberhauser:
- Moved composables that were in `AppState` to `MainActivityScreen`
- Moved anonymous objects that handled navigation callbacks from `MozillaNavHost` to `AppState`
- Renamed all `XRoute` composables to `XScreen`


app/src/main/java/org/mozilla/social/ui/MainActivityScreen.kt
```diff
@@ -1,26 +1,74 @@
...
 
+    val currentDestination = appState.currentDestination.collectAsState().value
+
     MoSoScaffold(
         modifier = Modifier.nestedScroll(appState.topAppBarScrollBehavior.nestedScrollConnection),
         snackbarHost = { appState.snackbarHostState },
-        floatingActionButton = { appState.FloatingActionButton() },
-        bottomBar = { appState.BottomBar() },
-        topBar = { appState.TopBar() },
+        floatingActionButton = {
```
devotaaabel, line 53: Curious why you moved these? I thought it was a nice way to encapsulate the navigation-specific logic in the app state

feature/followers/src/main/java/org/mozilla/social/feature/followers/FollowersScreen.kt
```diff
@@ -33,7 +33,7 @@ import org.mozilla.social.core.ui.pullrefresh.pullRefresh
 import org.mozilla.social.core.ui.pullrefresh.rememberPullRefreshState
 
 @Composable
-internal fun FollowersRoute(
+internal fun FollowersScreen(
```
devotaaabel, line 36: Why are these internal? As the main top-level composable for the module, IMO both the stateful and stateless composables should be public

app/src/main/java/org/mozilla/social/ui/MainActivityScreen.kt
JohnOberhauser: I think `AppState` was doing too much.  It was violating the single responsibility principal by housing UI code and navigation stuff, etc.  It feels like the equivalent of putting composables in a view model.

feature/followers/src/main/java/org/mozilla/social/feature/followers/FollowersScreen.kt
JohnOberhauser: These will either be internal or public based on the outcomes of this decision doc: https://docs.google.com/document/d/1S7DtfNz2VcUCXtLYrVDMN4zu_YDfWWvjn4EHZJMAubE/edit#heading=h.6rrsie4fel2

Should be find to have them be internal until we come to a decision.

app/src/main/java/org/mozilla/social/ui/MainActivityScreen.kt
devotaaabel: That's fine and reasonable, though this is changing the responsibility of `AppState` from: handling navigation logic and managing navigation-based UI state to -> handling navigation logic and serving up navigation state to the rest of the app

app/src/main/java/org/mozilla/social/ui/AppState.kt
```diff
@@ -100,10 +90,11 @@ class AppState(
     val coroutineScope: CoroutineScope,
     val bottomSheetVisible: MutableState<Boolean>,
     val snackbarHostState: SnackbarHostState,
+    val context: Context,
 ) {
 
     @OptIn(ExperimentalCoroutinesApi::class)
-    private val currentDestination: StateFlow<NavDestination?> =
+    val currentDestination: StateFlow<NavDestination?> =
```
devotaaabel, line 97: @JohnOberhauser this is based off of the nav destinations on line 241 being up to date (which they are not). We need an enforcement mechanism to ensure this is accurate if we're going to surface this throughout the app

JohnOberhauser: Conversing about this in slack

devotaaabel (approved):
Thanks for pairing on this!

PR #175, opened 2023-10-06
Profile updates

JohnOberhauser:
- Updating some UI stuff in the profile screens
- Making the user's bio expandable
- Went back to using `Layout` for the header images.  When I was doing size changes of the profile avatar, it made it so the avatar was not centered on the baseline of the header image.  Using `Layout` makes it so we can change the size without having to manually calculate the padding.


timc-mozilla (approved):

config/detekt/detekt.yml
```diff
@@ -4,6 +4,8 @@ complexity:
     # Relaxing this to a smaller degree is recommended for Compose anyway.
   TooManyFunctions:
     active: false # Carried over from Pocket configuration.
+  LongMethod:
+    threshold: 100 # Increasing from 60 to 100 because I find compose methods can run long sometimes.
```
devotaaabel, line 8: Agreed that this is a rule that is a little challenging to follow stringently and probably not always necessary to adhere to 👍

core/common/src/main/java/org/mozilla/social/common/utils/DateTimeFormatters.kt
```diff
@@ -0,0 +1,7 @@
+package org.mozilla.social.common.utils
+
+import java.time.format.DateTimeFormatter
+
+object DateTimeFormatters {
+    val standard: DateTimeFormatter = DateTimeFormatter.ofPattern("MMMM dd, yyyy")
```
devotaaabel, line 6: @JohnOberhauser I don't think this is actually used anywhere, but just FYI we'll need to use locale-specific date/time formatters

core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoTheme.kt
```diff
@@ -312,3 +315,15 @@ fun ProvideMoSoColors(
 private val localMoSoColors = staticCompositionLocalOf<MoSoColors> {
     error("No MoSo provided")
 }
+
+@Composable
+fun ProvideMoSoTextStyle(
+    content: @Composable () -> Unit,
+) {
+    CompositionLocalProvider(
+        LocalTextStyle provides MoSoTheme.typography.bodyMedium.copy(
```
devotaaabel, line 324: @JohnOberhauser  I don't think we only want to do this for body medium, I think we need to use [composition local](https://developer.android.com/jetpack/compose/compositionlocal#intro)/[local content color](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#LocalContentColor())

JohnOberhauser: Are you specifically talking about the color part here?  I think `LocalContentColor` will affect more than just text.  For this, I just wanted to set a default typography with a default text color.  `LocalTextStyle` provides both of those.  

If that doesn't make sense, I'd like to align on this over zoom

core/common/src/main/java/org/mozilla/social/common/utils/DateTimeFormatters.kt
JohnOberhauser: It's used in `AccountScreen` where it says the user's joined date.   

Good call on locale-specifics.  I'll create a ticket for that

PR #176, opened 2023-10-09
build: setup detekt

marcin-kozinski:
## Implementation decisions
* Added a Detekt convention plugin and applied it in our android app/library convention plugins.
  * This means Detekt is enabled for all modules by default. If we ever don't want it enabled for a module, we can look into providing an opt out.
  * This means Detekt runs for each module separately, not as a single task that analyses all Kotlin files in the repository. You can still run `./gradlew detekt` to run it for all modules in one command.
  * This means each module gets its own baseline file.
* I used the `buildUponDefaultConfig` option to keep our config minimal, easier to maintain as rules are deprecated/migrated in future versions and so it's clearer where we use default settings and where we customise.
* I created the config by merging Detekt's own recommendations for Compose projects and some customisations we used at Pocket. This is very much open to further tweaking, either as part of this review or future PRs.
* Since our GHA workflows started duplicating common setup steps a lot, I learned about custom composite actions and added one to solve this.

## References
* https://github.com/detekt/detekt
* https://detekt.dev/docs/gettingstarted/gradle
* https://github.com/Pocket/Android/pull/1718
* https://github.com/Pocket/Android/blob/e86118c98d6f9bcd0e2a19ad0eee9478f80fde17/config/detekt/detekt.yml

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -1,0 +1,25 @@
+name: Setup environment
+description: Common environment setup steps for all workflows
+inputs:
+  jdk:
+    description: Set up JDK
+    required: false
+    default: 'true'
+  gradle:
+    description: Set up Gradle
+    required: false
+    default: 'true'
+runs:
+  using: composite
+  steps:
+  - name: Set up JDK
+    if: ${{ inputs.jdk || inputs.gradle }}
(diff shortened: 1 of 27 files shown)


JohnOberhauser (approved):

devotaaabel (approved):
Yay 🎉

PR #177, opened 2023-10-10
chore(deps): clean up version catalog and applying plugins

marcin-kozinski:
Wanted to just reformat `libs.versions.toml` according to our recently introduced [rules](https://github.com/MozillaSocial/mozilla-social-android/blob/main/CONTRIBUTING.md#dependencies). While I was there I figured out a cleaner way to apply the plugins added for Glean.


JohnOberhauser (approved):
:tada:

PR #188, opened 2023-10-14
followers proguard fix

JohnOberhauser:
Followers and Following lists were breaking due to a proguard issue.

This retrofit proguard line is merged into retrofit's master branch but it's not released yet
https://github.com/square/retrofit/pull/3910/files/9831d5c3f9d28c61d29078d89645bce8ea258366#diff-50d428633d98e235831f5f9b75d7aa48897d5edc2bed30c55c9bb9ec20b36f82


marcin-kozinski (approved):
I left some cleanup/maintenance suggestions, but this is a good and needed fix in the current for too, so approving either way.

timc-mozilla (approved):

PR #189, opened 2023-10-14
Simplifying colors

JohnOberhauser:
@devotaaabel I think you were just copying what firefox did for their theme colors, right?  It seems a little complex to me, and I'm trying to figure out if there is any benefits to it.  When I was adding some colors to the other day, I found it tedious having to update the constructor, the variables, the update method, and the copy method.  Any reason we can't simplify it like so?

### DIFF
--- a/core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoTheme.kt
+++ b/core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoTheme.kt
@@ -14,10 +14,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.Stable
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.toArgb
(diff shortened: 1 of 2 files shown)


timc-mozilla (approved):
looks cleaner :D

core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoTheme.kt
```diff
@@ -89,6 +85,7 @@ private val darkColorPalette = MoSoColors(
     iconAccent = FirefoxColor.Violet30,
     iconActionActive = FirefoxColor.Violet20,
     iconActionDisabled = FirefoxColor.LightGrey70,
+    iconAccentViolet = FirefoxColor.Violet30,
```
devotaaabel, line 88: is this supposed to have "violet" in the name? Is it from the design system? I didn't see it in the figma though I could've been looking in the wrong place. Seems weird that it has the word "violet" in the name, as no other design system tokens do that.

devotaaabel (commented):
@JohnOberhauser re: your question about why- this is how material themes do it- curious if you looked into that/why they do it this way?

core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoTheme.kt
JohnOberhauser: I'm not sure.  This was previously on line 245.  I'm not sure how it ended up there.  I removed it and replaced it's usage.

devotaaabel: Ohh I see now, I'm pretty sure it was a typo

devotaaabel (approved):
Cool thanks for looking into it

PR #192, opened 2023-10-16
Report updates part 3

JohnOberhauser:
- Finishing the second screen in the flow where the user does the actual reporting
- Fixing small bugs in other areas of the flow

### DIFF
--- a/core/ui/src/main/java/org/mozilla/social/core/ui/loading/GenericLoading.kt
+++ b/core/ui/src/main/java/org/mozilla/social/core/ui/loading/GenericLoading.kt
@@ -1,0 +1,20 @@
+package org.mozilla.social.core.ui.loading
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import org.mozilla.social.core.designsystem.component.MoSoCircularProgressIndicator
+
+@Composable
+fun GenericLoading() {
+    Box(
+        modifier = Modifier
+            .fillMaxSize()
+            .wrapContentSize(Alignment.Center)
+    ) {
+        MoSoCircularProgressIndicator()
+    }
+}
--- a/feature/report/src/main/java/org/mozilla/social/feature/report/step1/ReportScreen1.kt
+++ b/feature/report/src/main/java/org/mozilla/social/feature/report/step1/ReportScreen1.kt
@@ -343,8 +343,9 @@ private fun AdditionalReportFields(
                 reportInteractions = reportInteractions,
             )
             Spacer(modifier = Modifier.height(16.dp))
-            externalInstance.value?.let {
+            externalInstance.value?.let { externalInstance ->
                 SendToOtherServerOption(
+                    externalInstance = externalInstance,
                     checked = sendToExternalServer,
                     reportInteractions = reportInteractions,
                 )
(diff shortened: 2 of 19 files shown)


feature/report/src/main/java/org/mozilla/social/feature/report/step2/ReportScreen2ViewModel.kt
```diff
@@ -29,28 +36,92 @@ class ReportScreen2ViewModel(
...
+
+    private val _reportIsSending = MutableStateFlow(false)
+    val reportIsSending = _reportIsSending.asStateFlow()
+
+    init {
+        getStatuses()
+    }
+
+    private fun getStatuses() {
+        _statuses.edit { Resource.Loading() }
```
devotaaabel, line 50: I'd prefer we use `update` instead of `edit` for these

core/ui/src/main/java/org/mozilla/social/core/ui/loading/GenericLoading.kt
```diff
@@ -0,0 +1,20 @@
...
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import org.mozilla.social.core.designsystem.component.MoSoCircularProgressIndicator
+
+@Composable
+fun GenericLoading() {
```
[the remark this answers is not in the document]
devotaaabel: Or just make it private where it's used, since it seems like a pretty simple and possibly specific impl.

feature/report/src/main/java/org/mozilla/social/feature/report/step2/ReportScreen2ViewModel.kt
```diff
@@ -29,28 +36,92 @@ class ReportScreen2ViewModel(
...
     override fun onCloseClicked() {
         onClose()
     }
+
+    override fun onStatusClicked(statusId: String) {
+        _statuses.edit {
+            Resource.Loaded(
+                data = (statuses.value as Resource.Loaded).data.map {
+                    if (it.statusId == statusId) {
+                        it.copy(
```
devotaaabel, line 113: IMO whether a checkbox is checked is UI-level logic and shouldn't need to go through the viewmodel (then the callback that's invoked when the form is submitted could just pass in the ids for the statuses that are selected)

devotaaabel (approved):

feature/report/src/main/java/org/mozilla/social/feature/report/step1/ReportScreen1.kt
```diff
@@ -129,7 +129,7 @@ private fun MainContent(
 ) {
     Column(
         modifier = Modifier
-            .padding(start = 16.dp, end = 16.dp)
+            .padding(horizontal = 16.dp)
```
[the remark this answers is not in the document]
JohnOberhauser: hmm, I'm a little wary of this `MoSoSpacing`.  I'm going to add a topic for the next TIP to discuss

feature/report/src/main/java/org/mozilla/social/feature/report/step2/ReportScreen2ViewModel.kt
JohnOberhauser: I tried doing this and it gets pretty messy in compose because it requires the button to know about the state of all the checked boxes.  It also requires an individual `SelectableStatusCard` to have access to the `MutableState<List<String>>`  that represents which boxes are checked (because the card then needs to be able to edit that list).

PR #193, opened 2023-10-16
Retain scroll

JohnOberhauser:
When navigating back to a list, the `lazyPagingItems` seem to have a list size of 0 for a split second before going back to where it was.  This causes the list scroll state to reset at 0, losing the scroll position.  I'm creating a second scroll state to use when the list is not 0... I'm not super confident in this solution, but it does work.  There is probably a better way though.  Ideally, the flow would not emit an empty list initially.  I'm not sure we can prevent it though.

- Retaining scroll position by providing two scroll states
- Updating top level navigation so a new feed isn't created every time we navigate to the feed


core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCardList.kt
```diff
@@ -74,9 +75,22 @@ fun PostCardList(
...
+        // to reset at 0, losing the scroll position.  The realState variable fixes this by
+        // only being in use when the item count is greater than 0.
+        // There might a better solution for this...
+        val emptyListState = rememberLazyListState()
+        val realState = rememberLazyListState()
+
         LazyColumn(
             Modifier
                 .fillMaxSize(),
+            state = if (lazyingPagingItems.itemCount == 0) {
```
devotaaabel, line 89: @JohnOberhauser I can't repro this, but does the list scroll to the right place, it's just you're trying to avoid the initial empty state from rendering? What about wrapping the timeline list in `AnimatedVisibility(visibility = items.itemCount != 0)`?

JohnOberhauser: No, the list scrolls to the top because the item count is briefly 0.

To reproduce, go multiple screens deep.  Like click a status to open a thread, then click a user's icon to open their profile from within the thread.  Then go back all the way to the list.  The scroll position should be lost.

JohnOberhauser: I've thought about hiding the `LazyColumn`, but there is a problem; from my understanding, `LazyColumn` should not be nested with other scrollable content.  The `PostCardList` takes a `header` content parameter for things like the account screen's content, which is above the user's timeline.  That content needs to always show, and it's inside an `item` in the `LazyColumn`; so we can't hide the `LazyColumn` ever.  Maybe there is a way to change all this?  idk

devotaaabel (approved):
Thanks for trying to find a sensible solution here 😭 also I think this is the relevant open bug ticket, it's status is "assigned": https://issuetracker.google.com/issues/179397301

PR #200, opened 2023-10-19
fixing issues with polls

JohnOberhauser:
- Fixing issue with voting request
- Making it so results show if the poll is expired
- Fixing issue where we were trying to write to database from the main thread

### DIFF
--- a/core/ui/src/main/java/org/mozilla/social/core/ui/poll/PollMapping.kt
+++ b/core/ui/src/main/java/org/mozilla/social/core/ui/poll/PollMapping.kt
@@ -37,7 +37,7 @@ fun Poll.toPollUiState(
                 )
             } else StringFactory.literal("")
         ),
-        showResults = hasVoted ?: false,
+        showResults = hasVoted ?: false || isExpired,
         pollId = pollId,
         isMultipleChoice = allowsMultipleChoices,
         usersVotes = ownVotes ?: emptyList(),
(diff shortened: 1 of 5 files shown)


core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCardDelegate.kt
```diff
@@ -22,7 +23,7 @@ class PostCardDelegate(
     val errorToastMessage = _errorToastMessage.asSharedFlow()
 
     override fun onVoteClicked(pollId: String, choices: List<Int>) {
-        coroutineScope.launch {
+        coroutineScope.launch(Dispatchers.IO) {
```
devotaaabel, line 26: We should get in the habit of injecting dispatchers for unit testing (for now just adding it to the constructor with a default param is fine). We should get in the habit of unit testing too....

core/ui/src/main/java/org/mozilla/social/core/ui/poll/PollMapping.kt
```diff
@@ -37,7 +37,7 @@ fun Poll.toPollUiState(
                 )
             } else StringFactory.literal("")
         ),
-        showResults = hasVoted ?: false,
+        showResults = hasVoted ?: false || isExpired,
```
devotaaabel, line 40:
```suggestion
        showResults = hasVoted ?: isExpired,
```

devotaaabel (approved):

core/ui/src/main/java/org/mozilla/social/core/ui/poll/PollMapping.kt
JohnOberhauser: I think this would not work if `hasVoted == false` and `isExpired == true`

PR #201, opened 2023-10-19
Created MoSoSnackbar

devotaaabel:
I tried to make this as additive as possible by using wrappers, so we don't have to maintain a whole custom snackbar implementation.

[snackbar.webm](https://github.com/MozillaSocial/mozilla-social-android/assets/14295757/bfef0989-b29f-4729-a0dd-0f6f4ce37c88)


JohnOberhauser (approved):
Looks great!  We should update `MoSoToast` to use this instead of the android toast

PR #202, opened 2023-10-19
Gradle cleanup

JohnOberhauser:
- Moving compose dependencies to the compose convention plugin
- Moving common android library setup to the android library convention plugin
- Removing unused dependencies from gradle files


devotaaabel (approved):
Nice 🔥 thanks for doing this

PR #206, opened 2023-10-23
[MOSOAND-164] new post UX updates

devotaaabel:
Updated new post flow so that new posts are automatically shown, and so users also see success snackbar when the post is successfully posted.


timc-mozilla (approved):

app/src/main/java/org/mozilla/social/navigation/MozillaNavHost.kt
```diff
@@ -67,7 +74,15 @@ private fun NavGraphBuilder.mainGraph(
         followersScreen(followersNavigationCallbacks = appState.followersNavigation)
         followingScreen(followersNavigationCallbacks = appState.followersNavigation)
         newPostScreen(
-            onStatusPosted = { appState.popBackStack() },
+            onStatusPosted = {
+                appState.popBackStack()
+                GlobalScope.launch {
+                    appState.snackbarHostState.showSnackbar(
```
JohnOberhauser, line 80: Is it possible to access the `snackbarHostState` from the feature modules, or will we need to use callbacks all the time?

JohnOberhauser (approved):

app/src/main/java/org/mozilla/social/navigation/MozillaNavHost.kt
devotaaabel: @JohnOberhauser This is what I'm going to use as an example of how we can avoid callbacks. So the idea is that we'll have a use case that can trigger snackbar events from anywhere

PR #207, opened 2023-10-24
[MOSOAND-106, MOSOAND-107] Added separate photo and video adding buttons

devotaaabel:
- Pulled BottomBar into it's own file and added BottomBarState
- There seems to be a bug in the compose dropdown preventing the original plan of showing these option in a dropdown, so I'm just showing separate icons until I have a chance to sync up with design
- Alphabetized MoSoIcons


feature/post/src/main/java/org/mozilla/social/post/bottombar/BottomBar.kt
```diff
@@ -0,0 +1,214 @@
...
+internal fun BottomBar(
+    bottomBarState: BottomBarState,
+    pollInteractions: PollInteractions,
+    contentWarningInteractions: ContentWarningInteractions,
+    onMediaInserted: (Uri, File, FileType) -> Unit,
+) {
+    val context = LocalContext.current
+    val multipleMediaLauncher = rememberLauncherForActivityResult(
+        ActivityResultContracts.PickMultipleVisualMedia(
+            maxItems = NewPostViewModel.MAX_IMAGES,
```
JohnOberhauser, line 49: You should be able to do `maxItems = (NewPostViewModel.MAX_IMAGES - currentItemCount).coerceAtLeast(2)` to make this dynamic.  You will need to use the `singleMediaLauncher` if the user has already uploaded 3 images, because the `mulitpleMediaLauncher` will required a max of at least 2.

devotaaabel: @JohnOberhauser since this is using `remember`, it doesn't get reinitialized on composition- it only gets called once. So it won't update based on new values when `currentNumberOfImagesAlreadyUploaded` changes

JohnOberhauser: I think it does.  `rememberLauncherForActivityResult` is not a normal `remember` function.  In fact, I wonder if we should `remember` the `ActivityResultContract`, since I think that might be getting instantiated on every recomposition.

devotaaabel: You're right, I think the existing `rememberLauncherForActivityResult` function does the right thing, I don't think we need another `remember` AFAIK

JohnOberhauser (approved):

PR #209, opened 2023-10-24
Discover part 3

JohnOberhauser:
- Updating endpoint we use to the new content endpoint
- Adding UI for the discover screen
- Making articles clickable
- Making share button work

### DIFF
--- a/core/data/src/main/java/org/mozilla/social/core/data/repository/RecommendationRepository.kt
+++ b/core/data/src/main/java/org/mozilla/social/core/data/repository/RecommendationRepository.kt
@@ -11,6 +11,6 @@ class RecommendationRepository(
 
     suspend fun getRecommendations(): List<Recommendation> =
         recommendationApi.getRecommendations(
-            locale = "${Locale.getDefault().language}-${Locale.getDefault().country}",
+            locale = Locale.getDefault().toLanguageTag(),
         ).recommendations.map { it.toExternalModel() }
 }
--- a/core/data/src/main/java/org/mozilla/social/core/data/repository/RecommendationRepository.kt
+++ b/core/data/src/main/java/org/mozilla/social/core/data/repository/RecommendationRepository.kt
@@ -1,6 +1,5 @@
 package org.mozilla.social.core.data.repository
 
-import org.mozilla.social.core.data.BuildConfig
 import org.mozilla.social.core.data.repository.model.recommendations.toExternalModel
 import org.mozilla.social.core.network.RecommendationApi
 import org.mozilla.social.model.Recommendation
@@ -12,7 +11,6 @@ class RecommendationRepository(
 
     suspend fun getRecommendations(): List<Recommendation> =
         recommendationApi.getRecommendations(
-            locale = Locale.getDefault().language,
-            consumerKey = BuildConfig.newTabConsumerKey,
+            locale = "${Locale.getDefault().language}-${Locale.getDefault().country}",
         ).recommendations.map { it.toExternalModel() }
 }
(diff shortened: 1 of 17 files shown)


core/data/src/main/java/org/mozilla/social/core/data/repository/RecommendationRepository.kt
```diff
@@ -12,7 +11,6 @@ class RecommendationRepository(
 
     suspend fun getRecommendations(): List<Recommendation> =
         recommendationApi.getRecommendations(
-            locale = Locale.getDefault().language,
-            consumerKey = BuildConfig.newTabConsumerKey,
+            locale = "${Locale.getDefault().language}-${Locale.getDefault().country}",
```
devotaaabel, line 14: This could be a useful extension function

core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoRadius.kt
```diff
@@ -7,4 +7,7 @@ object MoSoRadius {
     val md = 8.dp
     val lg = 16.dp
     val full = 360.dp
+
+    // radius for image and video previews
+    val media = 16.dp
```
devotaaabel, line 12: What do you think about moving this into a separate set of constants somewhere other than the design system module, so we know which values are associated with the design system?

feature/discover/src/main/java/org/mozilla/social/feature/discover/DiscoverScreen.kt
```diff
@@ -1,44 +1,177 @@
...
+                    CustomTabsIntent
+                        .Builder()
+                        .build()
+                        .launchUrl(
+                            context,
+                            recommendation.url.toUri(),
+                        )
+                }
+        ) {
+            Row {
```
devotaaabel, line 121: I think these multiple text boxes would be a good candidate for [grouping elements for accessibility](https://developer.android.com/jetpack/compose/accessibility#group-elements)

feature/discover/src/main/java/org/mozilla/social/feature/discover/DiscoverScreen.kt
```diff
@@ -1,44 +1,177 @@
...
+                        style = MoSoTheme.typography.bodyMedium,
+                    )
+                }
+                AsyncImage(
+                    modifier = Modifier
+                        .padding(start = 16.dp)
+                        .weight(1f)
+                        .clip(RoundedCornerShape(MoSoRadius.media)),
+                    model = recommendation.image.firstOrNull()?.url,
+                    contentDescription = recommendation.title,
```
devotaaabel, line 147: Does the API not return content descriptions for images? I think that would be the proper thing to do here. I think we'd rather this be null, considering a screen reader would have already read the title

devotaaabel (approved):
Looks good, I just added a few suggestions/comments

core/designsystem/src/main/java/org/mozilla/social/core/designsystem/theme/MoSoRadius.kt
JohnOberhauser: Moving this to the `ui` module in an object called `Radius`

core/data/src/main/java/org/mozilla/social/core/data/repository/RecommendationRepository.kt
JohnOberhauser: changing this to `Locale.getDefault().toLanguageTag()`

PR #210, opened 2023-10-24
glean cleanup

JohnOberhauser:
Glean was still trying to do stuff according to logcat, so I made some changes

- Separating `Analytics` and `GleanAnalytics` into different files
- Un-commenting out glean code in `GleanAnalytics`
- Making the `AnalyticsModule` provide a dummy `Analytics` object instead of the glean one

### DIFF
--- a/core/analytics/src/main/java/org/mozilla/social/core/analytics/AnalyticsModule.kt
+++ b/core/analytics/src/main/java/org/mozilla/social/core/analytics/AnalyticsModule.kt
@@ -1,0 +1,12 @@
+package org.mozilla.social.core.analytics
+
+import org.koin.dsl.module
+import org.mozilla.social.core.analytics.glean.GleanAnalytics
+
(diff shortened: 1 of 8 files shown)


core/analytics/src/main/java/org/mozilla/social/core/analytics/AnalyticsModule.kt
```diff
@@ -0,0 +1,12 @@
+package org.mozilla.social.core.analytics
+
+import org.koin.dsl.module
+import org.mozilla.social.core.analytics.glean.GleanAnalytics
+
+val analyticsModule = module {
+    single<Analytics> {
+        object : Analytics {}
```
devotaaabel, line 8: It would be nice if this could log to logcat on debug builds. I wonder if there's a way we could use reflection so we don't have to add an implementation every time we add a new event

devotaaabel (approved):

PR #211, opened 2023-10-24
Preventing clicking on root status in thread

JohnOberhauser:
- Making it so user's cannot click on a thread's root status


devotaaabel (commented):
@JohnOberhauser you can still click into a thread for statuses that aren't threads

JohnOberhauser:
> @JohnOberhauser you can still click into a thread for statuses that aren't threads

You will have to show me what you mean

devotaaabel (approved):

PR #212, opened 2023-10-24
Adding labels to profile

JohnOberhauser:
- adding labels to the profile in the bio section
- Not adding label specific icons right now.  That needs to be talked about with PM's and designers, and the prioritized

### DIFF
--- a/feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
+++ b/feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
@@ -46,11 +46,13 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.rotate
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.painter.Painter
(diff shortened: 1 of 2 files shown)


feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
```diff
@@ -544,52 +539,47 @@ private fun UserBio(
...
+        }
+
+        HtmlContent(
+            modifier = Modifier
+                .weight(1f),
+            mentions = emptyList(),
+            htmlText = text,
+            htmlContentInteractions = htmlContentInteractions,
+            textStyle = MoSoTheme.typography.bodySmall,
+            textColor = MoSoTheme.colors.textSecondary,
```
devotaaabel, line 579: Should text and link colors be the same?

devotaaabel (approved):

feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
JohnOberhauser: They are in the designs

PR #215, opened 2023-10-25
Screen transition animations

JohnOberhauser:
- Adding screen transition animations
  - There is a small amount of animation jank due to the top and bottom app bars in the scaffold.  Is there a way we can make the scaffold only exist on the main screens and not be a hidden element on all other screens?


timc-mozilla (approved):
not sure about the scaffold, but can we add negative padding to the appbar to have it be "above" the visible screen?

PR #216, opened 2023-10-26
Preview card

JohnOberhauser:
- Adding a link preview card to statuses
  - Had to update our models and mappings
  - Added the PreviewCard ui
- Reducing the length of links based on the html span tags


core/data/src/main/java/org/mozilla/social/core/data/repository/model/status/NetworkStatusToExternal.kt
```diff
@@ -226,3 +227,68 @@ fun NetworkSource.toExternalModel(): Source =
...
+            providerName = providerName,
+            providerUrl = providerUrl,
+            html = html,
+            width = width,
+            height = height,
+            image = image,
+            embedUrl = embedUrl,
+            blurHash = blurHash
+        )
+        else -> throw Exception("type value is incorrect")
```
devotaaabel, line 293: I like using `error("message")` for these types of things which defaults to an `IllegalStateException` which I think is appropriate here (also it would mean we don't have to supress the detekt rule)

core/ui/src/test/java/org/mozilla/social/core/ui/htmlcontent/ReduceHtmlLinksTest.kt
```diff
@@ -0,0 +1,35 @@
+package org.mozilla.social.core.ui.htmlcontent
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@Suppress("MaxLineLength")
+class ReduceHtmlLinksTest {
```
devotaaabel, line 7: 👏

devotaaabel (approved):

PR #218, opened 2023-10-27
Media aspect ratio

JohnOberhauser:
- Ensuring all media has the proper aspect ratio set so images, gifs, and videos don't even "pop in" when scrolling up
- Updated models to have the proper fields returned by the API


devotaaabel (approved):
Looks good 🔥

PR #219, opened 2023-10-27
Moved settings to profile overflow, moved app bar out of scaffold and updated animations

devotaaabel:
Moved settings to profile overflow and removed navigation drawer. Moved top app bar to feed screen and fixed window insets everywhere. Updated animations.


app/src/main/java/org/mozilla/social/ui/MainActivityScreen.kt
```diff
@@ -94,72 +88,21 @@ fun MainActivityScreen(context: Context) {
...
-            val padding = if (AppState.shouldShowTopBar(currentDestination)) {
-                paddingValues
-            } else {
-                WindowInsets.statusBars.asPaddingValues()
-            }
-            Box(modifier = Modifier.padding(padding)) {
-                MozillaNavHost(appState = appState, context = context)
-            }
+        content = {
+            MozillaNavHost(appState = appState, context = context)
```
JohnOberhauser, line 92: I'm guessing we still need to add the `paddingValues` from the lambda

JohnOberhauser: Or we could add bottom padding to screens that use the bottom bar, the same way you are doing the window insets padding

JohnOberhauser (approved):

PR #220, opened 2023-10-27
added settings and toggle analytics

timc-mozilla:
https://mozilla-hub.atlassian.net/browse/MOSOAND-135

Remade part of the settings screen
added Toggle Analytics
Defaulted to NOT uploading Glean data
Added new App Preferences Datastore

Kept DummyAnalytics, going to use it for testing I think


feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsViewModel.kt
```diff
@@ -20,8 +24,14 @@ class SettingsViewModel(
     private val _isToggled: MutableStateFlow<Boolean> = MutableStateFlow(false)
     var isToggled = _isToggled.asStateFlow()
 
+    private val _isAnalyticsToggledOn: MutableStateFlow<Boolean> = MutableStateFlow(false)
+    var isAnalyticsToggledOn = _isAnalyticsToggledOn.asStateFlow()
+
     init {
         viewModelScope.launch {
+            appPreferencesDatastore.trackAnalytics.collectLatest { toggled ->
```
JohnOberhauser, line 32: You will need to launch a new coroutine for this.  Since you are `collectLatest` is a blocking function, so everything past this block will not run

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsViewModel.kt
```diff
@@ -30,6 +40,16 @@ class SettingsViewModel(
         }
     }
 
+    fun toggleAnalytics() {
+        _isAnalyticsToggledOn.value = _isAnalyticsToggledOn.value.not()
```
JohnOberhauser, line 44: You don't need to update this value manually.  You are already collecting the flow from the data store, and that will update the toggle.

core/analytics/src/main/java/org/mozilla/social/core/analytics/glean/GleanAnalytics.kt
```diff
@@ -87,4 +87,8 @@ class GleanAnalytics : Analytics {
         Identifiers.mastodonAccountHandle.destroy()
         Identifiers.mastodonAccountId.destroy()
     }
+
+    override fun toggleAnalyticsTracking(toggleTracking: Boolean) {
```
JohnOberhauser, line 91: Instead of this function, can we do a `mapLatest` based on the preference saved in the datastore?  In the `initialize` function, I think you could add something like

```
GlobalScope.launch {
    appPreferencesDatastore.trackAnalytics.collectLatest { enabled ->
        Glean.setUploadEnabled(enabled)
    }
}
```

JohnOberhauser: That should make sure Glean gets enabled or disabled properly when restarting the app.  The way it's currently set up, I think it would always be off until the user uses the settings toggle for every app session.

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsScreen.kt
```diff
@@ -40,19 +40,226 @@ import org.mozilla.social.core.designsystem.theme.MoSoTheme
...
+            )
+        }
         Column(verticalArrangement = Arrangement.Bottom) {
             MoSoButtonSecondary(
                 modifier = Modifier
                     .wrapContentHeight()
                     .fillMaxWidth()
                     .padding(8.dp),
-                onClick = { settingsViewModel.onLogoutClicked() }
+                onClick = { logUserOut }
```
devotaaabel, line 77:
```suggestion
                onClick = logUserOut
```

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsScreen.kt
```diff
@@ -40,19 +40,196 @@ import org.mozilla.social.core.designsystem.theme.MoSoTheme
...
+                            Text(
+                                text = stringResource(id = subtitle),
+                                modifier = Modifier.padding(top = 8.dp),
+                                style = MaterialTheme.typography.bodySmall,
+                                textAlign = TextAlign.Start,
+                            )
+                        }
+                    }
+                    Spacer(modifier = Modifier.weight(1f))
+                    Switch(
```
JohnOberhauser, line 175: Can you create a `MoSoSwitch` component based on the design system?  https://www.figma.com/file/QquqdFHNTs8E5ZgW6VImGL/Moz-Social---Design-System-Library?type=design&node-id=402-11331&mode=design&t=p7uaFBdIDjvF60VG-0

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsViewModel.kt
```diff
@@ -1,13 +1,49 @@
...
+    private val analytics: Analytics,
     private val logout: Logout,
 ) : ViewModel() {
+    private val _isAnalyticsToggledOn: MutableStateFlow<Boolean> = MutableStateFlow(false)
+    var isAnalyticsToggledOn = _isAnalyticsToggledOn.asStateFlow()
 
-    fun onLogoutClicked() {
-        logout()
+    init {
+        GlobalScope.launch {
```
JohnOberhauser, line 23: You should use `viewModelScope` here instead.  `GlobalScope` will keep this running until the app is killed.  `viewModelScope` will only last as long as the view model

timc-mozilla: ok, i was just going off of what you said in an earlier post:

`GlobalScope.launch {
    appPreferencesDatastore.trackAnalytics.collectLatest { enabled ->
        Glean.setUploadEnabled(enabled)
    }
}`

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsViewModel.kt
```diff
@@ -1,13 +1,49 @@
...
+            appPreferencesDatastore.trackAnalytics.collectLatest { enabled ->
+                saveSettingsChanges(enabled)
```
JohnOberhauser, line 25: I don't think you should call `saveSettingsChanges` here.  It will trigger analytics and set the datastore value.  Instead just do `_isAnalyticsToggledOn.value = it`.  Then you can also remove this:
```
        appPreferencesDatastore.trackAnalytics.collectLatest {
            _isAnalyticsToggledOn.value = it
        }
```
from `saveSettingsChanges`

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsScreen.kt
JohnOberhauser: The design system components should go in the design system module in the components package.  There are a lot of other components we've already put together there, so you can use them as examples on how to do this.  We basically just wrap the material components and apply the proper colors

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsViewModel.kt
JohnOberhauser: No worries.  That one was in the `GleanAnalytics` class where we don't have a specific coroutine scope, and we want that one to run for the life of the app, so `GlobalScope` was a good fit

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsViewModel.kt
timc-mozilla: yup, got it

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsScreen.kt
timc-mozilla: gotcha, added in

PR #224, opened 2023-10-28
[MOSOAND-207] Added EventRelay to for high-level app events like navigation and snackbars

devotaaabel:
- Cleaned up some unused UI.
- Created `EventRelay`, for relaying navigation and other high-level app events like snackbars
- Callback implementations can be injected into composable functions which will emit navigation events via the `NavigateTo` use case, passing in a `NavDestination`. Every navigation destination and it's parameters will be represented by a type that extends `NavDestination`. See `PostCardNavigationImpl` for an example
- Separated bottom tabs into separate nav host and graph. So now in `AppState` there's `mainNavController` and `tabbedNavController`, which correspond to `MainNavHost` and `TabbedNavHost`. 
- Added `MosoCloseableAppBar` which automatically pops the backstack
- Todo (possibly in a follow up PR):
- [ ] Write tests
- [ ] Write documentation


app/src/main/java/org/mozilla/social/ui/PostCardNavigationImpl.kt
```diff
@@ -0,0 +1,35 @@
+package org.mozilla.social.ui
+
+import org.mozilla.social.core.navigation.NavDestination
+import org.mozilla.social.core.navigation.NavigateTo
+import org.mozilla.social.core.ui.postcard.PostCardNavigation
+
+class PostCardNavigationImpl(private val navigateTo: NavigateTo): PostCardNavigation {
```
JohnOberhauser, line 7: I don't think this interface is necessary anymore.  The methods only get called in `PostCardDelegate`.  I think you can replace `PostCardNavigation` in the constructor for `PostCardDelegate` with `NavigateTo`.  Then just call `navigateTo(NavDestination.NewPost(replyId = statusId))` instead of `postCardNavigation.onReplyClicked(statusId)`, etc

feature/account/src/main/java/org/mozilla/social/feature/account/AccountNavigationCallbacks.kt
```diff
@@ -1,11 +1,37 @@
 package org.mozilla.social.feature.account
 
+import org.mozilla.social.core.navigation.NavDestination
+import org.mozilla.social.core.navigation.NavigateTo
+
 interface AccountNavigationCallbacks {
```
JohnOberhauser, line 6: Same thing here.  This interface can be completely removed.  We just need to replace `AccountNavigationCallbacks` with `NavigateTo` in `AccountViewModel`

app/src/main/java/org/mozilla/social/navigation/MainNavHost.kt
```diff
@@ -112,21 +77,94 @@ private fun NavGraphBuilder.mainGraph(
...
+            navigateToLogin = navigateToLogin,
+            navigateToLoggedInGraph = navigateToLoggedInGraph
+        )
+
+    }
+}
+
+fun NavGraphBuilder.bottomTabScreen(appState: AppState) {
+    composable(
+        route = NavigationDestination.Feed.route,
```
JohnOberhauser, line 106: I think this route should not be the same as the feed route

app/src/main/java/org/mozilla/social/ui/NavigationViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.launch
+import org.mozilla.social.core.domain.IsSignedInFlow
+import org.mozilla.social.core.navigation.MoSoNavigationRelay
+import org.mozilla.social.core.navigation.NavDestination
+import org.mozilla.social.core.navigation.NavigateTo
+
+class NavigationViewModel(
```
JohnOberhauser, line 12: I'm wondering if we can just put the stuff in the `init` block in the `MainViewModel`; then just inject the navigation relay into the app state using `koinInject<MoSoNavigationRelay>()` in the `rememberAppState` and collect the flow in `AppState` directly instead of using the `LaunchedEffect` in `MainActivityScreen`.

app/src/main/java/org/mozilla/social/ui/AppState.kt
```diff
@@ -96,150 +122,58 @@ class AppState(
...
+                        reportAccountId = accountId,
+                        reportAccountHandle = accountHandle,
+                        reportStatusId = statusId,
+                    )
+                }
+
+                is NavDestination.Thread -> mainNavController.navigateToThread(threadStatusId = statusId)
+                is NavDestination.Following -> mainNavController.navigateToFollowing(accountId)
+                is NavDestination.MyAccount -> tabbedNavController.navigateToMyAccount()
+                is NavDestination.Account -> tabbedNavController.navigateToAccount(accountId = accountId)
```
JohnOberhauser, line 158: Getting a crash when going to a user's account

This should be `is NavDestination.Account -> mainNavController.navigateToAccount(accountId = accountId)`

app/src/main/java/org/mozilla/social/ui/AppState.kt
```diff
@@ -96,150 +122,58 @@ class AppState(
...
+                    mainNavController.navigateToReport(
+                        reportAccountId = accountId,
+                        reportAccountHandle = accountHandle,
+                        reportStatusId = statusId,
+                    )
+                }
+
+                is NavDestination.Thread -> mainNavController.navigateToThread(threadStatusId = statusId)
+                is NavDestination.Following -> mainNavController.navigateToFollowing(accountId)
+                is NavDestination.MyAccount -> tabbedNavController.navigateToMyAccount()
```
JohnOberhauser, line 157: Looks like `NavDestination.MyAccount` is the only tabbed destination here.  I think `NavDestination.MyAccount` can be removed, since the navigation to my account is handled in `AppState.navigateToBottomBarDestination`

JohnOberhauser:
When using the system back key on the home screen, the app goes to a blank page

devotaaabel:
> When using the system back key on the home screen, the app goes to a blank page
> 
> I'm not sure if this is the right solution, but you can add this to the Feed screen to fix it
> 
> ```
> val context = LocalContext.current
> BackHandler {
>     (context as? Activity)?.finish()
> }
> ```

I seem to have fixed this but lmk if you see it again

JohnOberhauser (approved):
:fire:

PR #226, opened 2023-10-31
Delete a post

JohnOberhauser:
- Adding ability to delete a post
- Making it so you can't report, mute, or block yourself
- Adding insert / delete animations to post card lists


core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCard.kt
```diff
@@ -192,18 +208,47 @@ private fun MetaData(
...
-                    overflowMenuExpanded.value = false
-                }
-            ) {
+        MoSoDropdownMenu(
+            expanded = overflowMenuExpanded.value,
+            onDismissRequest = {
+                overflowMenuExpanded.value = false
+            }
+        ) {
+            if (post.isUsersPost) {
```
devotaaabel, line 245: I think it would be a good pattern for options like this to be included in the PostCardUiState (could be a data class made of flags for each option). So that way we could pull the business logic out of the composable

devotaaabel (approved):

core/ui/src/main/java/org/mozilla/social/core/ui/postcard/PostCard.kt
JohnOberhauser: The `post` variable is here a `MainPostCardUiState` which is a variable inside `PostCardUiState`, so if I'm understanding your comment correctly, this should be good

PR #227, opened 2023-10-31
Edit profile part 3 - bot and lock

JohnOberhauser:
- Adding bot and lock checkboxes to the edit profile screen


core/designsystem/src/main/java/org/mozilla/social/core/designsystem/icon/MoSoIcons.kt
```diff
@@ -36,6 +36,9 @@ object MoSoIcons {
     @Composable
     fun bookmark() = painterResource(R.drawable.bookmark_simple)
 
+    @Composable
+    fun bot() = painterResource(id = R.drawable.bot)
```
devotaaabel, line 40: Do these names correspond with the phosphor icons? I know designs don't always have the right names listed, but I've been trying to keep the drawables and functions consistent with the names of the actual phosphor icons so that we're able to find them later. Then you can always set up `bot` as an alias, so something like 
```
 @Composable
    fun bot() = realIconName()

 @Composable
    fun realIconName() = painterResource(id = R.drawable.real_icon_name)

File: core/designsystem/src/main/java/org/mozilla/social/core/designsystem/icon/MoSoIcons.kt [hunk 2]
```diff
@@ -97,6 +100,9 @@ object MoSoIcons {
     @Composable
     fun x() = painterResource(id = R.drawable.x)
 
+    @Composable
+    fun profileLock() = painterResource(id = R.drawable.lock)
```
devotaaabel, line 104: Same here

devotaaabel (approved):

core/designsystem/src/main/java/org/mozilla/social/core/designsystem/icon/MoSoIcons.kt
JohnOberhauser: The phosphor website seems to be down at the moment.  I can double check these later

JohnOberhauser: nvm, it was my vpn.  I'll update these now

PR #229, opened 2023-11-01
Small fixes

JohnOberhauser:
Fixing a handful of things

- Fixing top app bar color
- Making a back arrow the default back button for the top app bar
- Setting the screen fade animation duration to 250.  100 was too fast
- Removing the slide up animation for the thread screen (@devotaaabel slide up felt odd to me here, lmk if this is okay with you)
- Fixing window inset padding on the discover and account screen.
- Adding a divider to the top of the bottom bar
- Fixing a bug with screen rotation

### DIFF
--- a/app/src/main/java/org/mozilla/social/ui/AppState.kt
+++ b/app/src/main/java/org/mozilla/social/ui/AppState.kt
@@ -88,10 +88,7 @@ class AppState(
         }
 
     init {
-        this.tabbedNavController = tabbedNavController
-
         coroutineScope.launch(Dispatchers.Main) {
-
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/social/ui/AppState.kt
```diff
@@ -88,10 +88,7 @@ class AppState(
         }
 
     init {
-        this.tabbedNavController = tabbedNavController
```
devotaaabel, line 91: Oops ty

devotaaabel (approved):
Nice :shipit:

PR #233, opened 2023-11-02
Follow count fix

JohnOberhauser:
- Making the database update your following count when you follow or unfollow someone
- Adding unit tests for these methods
- Creating a base test for repository testing

### DIFF
--- a/core/data/src/main/java/org/mozilla/social/core/data/repository/AccountRepository.kt
+++ b/core/data/src/main/java/org/mozilla/social/core/data/repository/AccountRepository.kt
@@ -1,8 +1,8 @@
 package org.mozilla.social.core.data.repository
 
+import androidx.room.withTransaction
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
-import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
 import okhttp3.MultipartBody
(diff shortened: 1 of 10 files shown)


core/data/src/main/java/org/mozilla/social/core/data/repository/AccountRepository.kt
```diff
@@ -105,23 +106,44 @@ class AccountRepository internal constructor(
     suspend fun getAccountFavourites(): List<Status> =
         accountApi.getAccountFavourites().map { it.toExternalModel() }
 
-    suspend fun followAccount(accountId: String) {
+    suspend fun followAccount(
```
devotaaabel, line 109: Would you be open to putting this logic in a use case (or at least a separate function)? My thinking is that we'd only want this to happen if a user is looking at some UI with the follow count. Otherwise it's kind of wasted cycles

JohnOberhauser: I think this logic should run even if you're not looking at the UI.  You might navigate back to a screen that is affected by this and expect the UI to be updated.  I'm worried about someone not calling this function when following an account, and creating a bug in the process.

I actually have a very related topic added to the TIP agenda.  Is it okay to merge this in as-is for now?  I might end up refactoring a lot of stuff into use-cases depending on the outcome of that conversation.

devotaaabel (approved):

PR #235, opened 2023-11-03
Edit profile part 6 - Custom fields

JohnOberhauser:
- Adding custom fields to edit profile


feature/account/src/main/res/values/strings.xml
```diff
@@ -29,4 +29,9 @@
     <string name="edit_account_save_failed">Failed to save changes</string>
     <string name="edit_account_lock">Lock</string>
     <string name="edit_account_bot">Bot</string>
+
+    <string name="edit_account_metadata_title">Profile metadata</string>
+    <string name="edit_account_metadata_description">You can have up to 4 items displayed as a table on your phone</string>
```
devotaaabel, line 34: Should this be "on your profile"? I imagine it shows on a computer too lol

devotaaabel (approved):

feature/account/src/main/res/values/strings.xml
JohnOberhauser: whoops, good catch

PR #236, opened 2023-11-03
fixing timelines

JohnOberhauser:
For some reason, I started getting 19 statuses back from the API during home timeline paging, even though I'm requesting 20.  Our current logic is that if the response count != the request page size, that means we are done paging.  We should instead be checking if there is a `next` link header.

- Updating all timelines to grab to look at the link header for determining if we should continue paging

### DIFF
--- a/core/data/src/main/java/org/mozilla/social/core/data/repository/TimelineRepository.kt
+++ b/core/data/src/main/java/org/mozilla/social/core/data/repository/TimelineRepository.kt
@@ -3,6 +3,8 @@ package org.mozilla.social.core.data.repository
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import org.mozilla.social.common.parseMastodonLinkHeader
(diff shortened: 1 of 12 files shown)


core/data/src/main/java/org/mozilla/social/core/data/repository/TimelineRepository.kt
```diff
@@ -22,12 +25,22 @@ class TimelineRepository internal constructor(
...
-        timelineApi.getHomeTimeline(
+    ): StatusPagingWrapper {
+        val response = timelineApi.getHomeTimeline(
             olderThanId = olderThanId,
             immediatelyNewerThanId = immediatelyNewerThanId,
             limit = loadSize,
-        ).map { it.toExternalModel() }
+        )
+
+        if (!response.isSuccessful) {
```
devotaaabel, line 35: I think ideally we'd return a type that can either be a success or error (obviously we're not doing that anywhere but I think that's the ideal state). Just something to aspire towards...

devotaaabel (approved):

PR #238, opened 2023-11-03
Updated navigation sealed classes to reflect nav graph

devotaaabel:
- Deleted `NavDestination`, moved most destinations into `NavigationDestination`, and also created `BottomBarNavigationDestination` and `ReportNavigationDestination` 
    - @JohnOberhauser lmk if you have any issues with this. I didn't really touch the logic of the report screen for fear of breaking something, I just moved it into this new sealed class. It seems like a good idea to have these separate so no one accidentally navigates to a sub-graph
- Deleted `searchScreen` since we're not going to have a separate screen
- Moved `navigateTo..` extension functions to the `NavigationDestination` implementations
- Added Truth fluent assertions dependency (lmk how you feel about this)
- Updated the `NavigateTo` use case to be able to navigate to bottom bar destinations


app/src/main/java/org/mozilla/social/ui/AppState.kt
```diff
@@ -145,61 +147,77 @@ class AppState(
...
-     */
-    fun navigateToBottomBarDestination(destination: NavigationDestination) {
+    private fun navigateToBottomBarDestination(destination: BottomBarNavigationDestination) {
         Timber.d("NAVIGATION navigate to bottom bar destination: $destination")
 
-        if (destination == NavigationDestination.NewPost) {
-            mainNavController.navigateToNewPost()
-            return
-        }
+//        if (destination == NavigationDestination.NewPost) {
```
JohnOberhauser, line 206: I think this can be removed

JohnOberhauser (approved):

JohnOberhauser:
> Deleted searchScreen since we're not going to have a separate screen

That's fine, but I was thinking the search icon might open a new screen.  Would be good to put the search feature in a separate module

app/src/main/java/org/mozilla/social/ui/AppState.kt
devotaaabel: I'll remove it in my next PR

devotaaabel:
> > Deleted searchScreen since we're not going to have a separate screen
> 
> That's fine, but I was thinking the search icon might open a new screen. Would be good to put the search feature in a separate module

Yea let's just add it when we make the module

PR #243, opened 2023-11-05
Added wrapper extension functions for call logging

devotaaabel:
I added these extension functions for wrapping network/db calls that are a little risky. Case in point- when logging out, we get some race conditions related to data that is getting deleted on logout. When wrapping these two calls, I get this output:

Screenshot 2023-11-05 at 5 01 06 PM

It would be nice to get the stack trace when it's lost. I think that'll take some investigation


devotaaabel:
Closing to open a new PR from a different branch

PR #246, opened 2023-11-06
Fixed crash on logout

devotaaabel:


feature/settings/src/main/java/org/mozilla/social/feature/settings/account/AccountSettingsViewModel.kt
```diff
@@ -2,18 +2,26 @@ package org.mozilla.social.feature.settings.account
...
+import timber.log.Timber
 
 class AccountSettingsViewModel(
     private val logout: Logout,
     getAccountFlow: AccountFlow,
 ) : ViewModel() {
 
-    val userHeader = getAccountFlow().mapLatest { it.toUserHeader() }
+    val userHeader = try {
+        getAccountFlow().mapLatest { it.toUserHeader() }
```
JohnOberhauser, line 20: would it work to do this instead?
```
val userHeader = getAccountFlow()
    .catch {
        Timber.e(it)
    }.mapLatest {
        it.toUserHeader()
    }
```

devotaaabel:
This was fixed in https://github.com/MozillaSocial/mozilla-social-android/pull/248

PR #250, opened 2023-11-06
Snackbar fixes

JohnOberhauser:
- Moving some files in the main app module around
- Increasing the buffer of `EventRelay` so events in quick succession always get sent
- Removing scaffold from the tab screen.  Just using a column instead
- Adding a snackbar host to the MainActivityScreen
  - Setting the proper paddings, including a new padding extension function for the bottom bar that only adds padding if we are on the Tabs destination
- removing `MoSoToast`
  - Making everything that was using this use the snackbar instead
- Fixing the snackbar UI so it matches the design system
  - Had to remove the material Snackbar because it didn't it when I messed with the height.  Replaced it with a surface.
- added `androidx.lifecycle:lifecycle-runtime-compose` to the compose convention plugins


app/src/main/java/org/mozilla/social/ui/MainActivityScreen.kt
```diff
@@ -1,15 +1,45 @@
...
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import org.mozilla.social.MainViewModel
+import org.mozilla.social.core.designsystem.component.MoSoSnackbar
+import org.mozilla.social.core.designsystem.component.MoSoSnackbarHost
 import org.mozilla.social.navigation.MainNavHost
+import org.mozilla.social.ui.bottombar.MoSoNavigationBarDefaults.bottomBarPadding
 
+/**
+ * must pass in view model... why does app break without it?
```
devotaaabel, line 19: `by viewModel()` in `MainActivity` is a lazy delegate so it only instantiates the view model once it's been accessed... There's also a `getViewModel` koin function that we can use that's not lazy but we'd have to be careful to instantiate it at the right time

core/ui/postcard/src/main/java/org/mozilla/social/core/ui/postcard/PostCardList.kt
```diff
@@ -254,5 +254,5 @@ fun PostCardList(
         }
     }
 
-    MoSoToast(toastMessage = errorToastMessage)
+    MoSoErrorToast(toastMessage = errorToastMessage)
```
devotaaabel, line 257: Do we even need the toast anymore?

devotaaabel (approved):

core/ui/postcard/src/main/java/org/mozilla/social/core/ui/postcard/PostCardList.kt
JohnOberhauser: Good point.  I will refactor the usages of this to just send the snackbar events in the view models

PR #253, opened 2023-11-08
Adding coroutine scope that cancels on logout

JohnOberhauser:
Useful reading for understanding this
https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad
https://developer.android.com/kotlin/coroutines/coroutines-best-practices#create-coroutines-data-layer

- Creating a singleton `appScope` that is a coroutine scope with a `SupervisorJob`
  - The `SupervisorJob` gets canceled on logout, and a new `SupervisorJob` takes it's place
- Moving a bunch of repository function to use cases
- For functions in the repository that we don't want to be canceled when the `viewModelScope` is canceled, we make the following changes
  - Within the dispatcher context, we launch a coroutine from `appScope` with `async` + `await`.  Using `async` + `await` allows the `viewmodelScope` to cancel immediately when the user navigates away from the screen.  If we use `launch` + `join`, then `viewmodelScope` doesn't cancel until after the `appScope` work is done.  Same with just using `withContext(appScope)`
  - The appscope coroutine context is set to `dispatcherIo` which is an injected dispatcher that defaults to `Dispatchers.IO`. 
  - The work is wrapped in a try catch that will rethrow any exception, but wrap it in a custom exception.  We specify the exception like this so we can catch the specific exception in the viewmodel.  We need to move away from catching `Exception` for these because that will also catch a `CancellationException`.  We don't want to catch a `CancellationException` because we don't want error toasts showing up just because the user navigated away from the screen.  A `CancellationException` will not crash the app if not caught.
- Updating unit tests


devotaaabel (commented):
Looks good

core/data/src/main/java/org/mozilla/social/core/data/RepositoryModule.kt
```diff
@@ -1,6 +1,7 @@
 package org.mozilla.social.core.data
 
 import org.koin.dsl.module
+import org.mozilla.social.common.appscope.AppScope
```
devotaaabel, line 4: 👾

core/domain/src/test/java/org/mozilla/social/core/domain/BaseDomainTest.kt
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Used for use cases that should continue running even if their outer scope gets cancelled
+     * @param delayedCallBlock a block of a call that should be slightly delayed.  For example,
+     * if your use case uses [AccountApi.followAccount], then this parameter should look like
+     * { accountApi.followAccount(any()) }
+     * @param subjectCallBlock a block that is calling you use case
+     * @param verifyBlock a block that we should verify gets called once.  Typically good to use
+     * the same call that you used in the @param delayedCallBlock.
+     */
+    protected fun testOuterScopeCancelled(
```
devotaaabel, line 98: Nice

devotaaabel (approved):
This is great 😍🔥 thanks for doing this

core/data/src/main/java/org/mozilla/social/core/data/RepositoryModule.kt
devotaaabel: Are we missing some ktlint rules?

core/common/src/main/java/org/mozilla/social/common/appscope/AppScope.kt
```diff
@@ -0,0 +1,19 @@
+package org.mozilla.social.common.appscope
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlin.coroutines.CoroutineContext
+
+class AppScope : CoroutineScope {
```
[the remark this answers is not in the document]
JohnOberhauser: I'm not sure this can replace `GlobalScope` everywhere.  There might be some `GlobalScope`s that we want to continue on logout

core/data/src/main/java/org/mozilla/social/core/data/RepositoryModule.kt
JohnOberhauser: I think so

PR #256, opened 2023-11-08
Analytics screen impressions

timc-mozilla:
screen impression analytics


devotaaabel (approved):
LGTM. If you want to see a pattern that could apply here for a way to avoid having to call into the viewmodel, take a look at how the use case `PopNavBackStack` is setup/used with injection in `MoSoCloseableTopAppBar`. I think it works well for simple one-shot events like this (I'm envisioning single-use use cases like `OnHashTagScreenViewed`, etc.)

timc-mozilla:
> LGTM. If you want to see a pattern that could apply here for a way to avoid having to call into the viewmodel, take a look at how the use case `PopNavBackStack` is setup/used with injection in `MoSoCloseableTopAppBar`. I think it works well for simple one-shot events like this (I'm envisioning single-use use cases like `OnHashTagScreenViewed`, etc.)

I'm not sure what you mean by this.
I'll throw something in your calendar to chat about what your suggesting.  As there are other merge conflicts, i can't merge it in right away, so I'll set a time for us tomorrow.

PR #257, opened 2023-11-08
Deleted risky use cases in favor of synchronous account id get

devotaaabel:
- Cleaned up gradle dependencies
- Deleted various use cases for getting the user account id and is signed in, in favor of renamed `LoggedInUserAccountId` use case which accesses the id synchronously according to the data store recommendations (using `first()`)
- Added data preloading to main activity per the [official recommendations](https://developer.android.com/topic/libraries/architecture/datastore#synchronous). In a future PR I might experiment with preloading after login or logout, but I think this is a good start


core/domain/src/main/java/org/mozilla/social/core/domain/GetLoggedInUserAccountId.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.coroutines.runBlocking
+import org.mozilla.social.core.datastore.UserPreferencesDatastore
+import timber.log.Timber
+
+/**
+ * The account ID of the logged in user
+ *
+ * So much depends on this value, trying out a blocking call to simplify the rest of the code.
+ *
+ * Uses a mutex so we can cancel the flow collection after we get the value.
```
JohnOberhauser, line 13: Can you update this comment?

gradle/libs.versions.toml
```diff
@@ -59,6 +60,7 @@ coil-gif = { module = "io.coil-kt:coil-gif", version.ref = "coil" }
 coil-video = { module = "io.coil-kt:coil-video", version.ref = "coil" }
 glean = { module = "org.mozilla.telemetry:glean", version.ref = "glean" }
 glean-forUnitTests = { module = "org.mozilla.telemetry:glean-native-forUnitTests", version.ref = "glean" }
+google-truth = { module = "com.google.truth:truth", version.ref = "google-truth" }
```
JohnOberhauser, line 63: I don't think we actually need this dependency.  There are a bunch of assertion function in `kotlin.test` already

devotaaabel: @JohnOberhauser Using [fluent assertions](https://truth.dev/) is just a personal preference- it just makes it easier/more idiomatic to make assertions. That link explains the differences

JohnOberhauser (approved):

PR #266, opened 2023-11-10
Modules - split up core/network module into core/network/mastodon and core/network/mozilla

devotaaabel:


core/data/src/main/java/org/mozilla/social/core/data/RepositoryModule.kt
```diff
@@ -23,5 +23,6 @@ fun repositoryModule(isDebug: Boolean) = module {
     single { RecommendationRepository(get()) }
     single { AppRepository(get()) }
     single { InstanceRepository(get()) }
-    includes(networkModule(isDebug))
+    includes(org.mozilla.social.core.network.mozilla.mozillaNetworkModule(isDebug))
```
JohnOberhauser, line 26: Can you add `org.mozilla.social.core.network.mozilla.mozillaNetworkModule` to imports?

JohnOberhauser: Looks like there are a lot of places that can be cleaned up like this

devotaaabel: I will in my next PR

JohnOberhauser (approved):

JohnOberhauser (approved):

PR #267, opened 2023-11-11
Login

JohnOberhauser:
I will have a part 2 for login, but I'm going to wait for the module changes to be finished before I start touching use-cases

- Updating login screen to reflect designs (portrait and landscape)
- Creating select server screen
- Adding navigation for login flow
- Adding some window size class functions

.

.


app/src/main/java/org/mozilla/social/MainViewModel.kt
```diff
@@ -26,7 +27,7 @@ class MainViewModel(
         viewModelScope.launch(Dispatchers.Main) {
             isSignedInFlow().collectLatest {
                 if (!it) {
-                    navigateTo(NavigationDestination.Login)
+                    navigateTo(AuthNavigationDestination.Login)
```
devotaaabel, line 30: Shouldn't the old value here be correct? That's the destination of the nested graph in the main graph

core/designsystem/src/main/res/font/zillaslab_bold.ttf
devotaaabel, line 1: Since when are we using custom fonts?

devotaaabel (commented):
I'll wait until opening another PR until this merges, please don't merge this until the other two are merged. I think it'll be a lot easier to resolve conflicts from your side if they arise

core/designsystem/src/main/res/font/zillaslab_bold.ttf
JohnOberhauser: This is just for the `mozilla.social` text on the login screen

app/src/main/java/org/mozilla/social/MainViewModel.kt
JohnOberhauser: oops, yeah

feature/auth/src/main/java/org/mozilla/social/feature/auth/chooseServer/ChooseServerViewModel.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.mozilla.social.core.usecase.mastodon.auth.Login
+
+class ChooseServerViewModel(
+    private val login: Login,
+) : ViewModel(), ChooseServerInteractions {
+
+    private val _uiState = MutableStateFlow(ChooseServerUiState())
+    val uiState = _uiState.asStateFlow()
+
+    @Suppress("SwallowedException")
```
devotaaabel, line 19: Is this supposed to be here?

feature/auth/src/main/java/org/mozilla/social/feature/auth/login/LoginScreen.kt
```diff
@@ -0,0 +1,224 @@
...
+import org.mozilla.social.core.designsystem.theme.MoSoSpacing
+import org.mozilla.social.core.designsystem.theme.MoSoTheme
+import org.mozilla.social.feature.auth.R
+
+@Composable
+internal fun LoginScreen(
+    viewModel: LoginViewModel = koinViewModel(),
+) {
+    val configuration = LocalConfiguration.current
+    if (configuration.orientation == Configuration.ORIENTATION_LANDSCAPE) {
```
devotaaabel, line 49: Can you use the compact size class here that was in the old login screen plz? That's supposed to be more flexible than `ORIENTATION_LANDSCAPE`

JohnOberhauser: I added a commit to create some window size class convenience functions.  I was going to switch to using window class as per your recommendation, but as I was deciding which window size classes should show which layout, I realized that using orientation is the right way to go in this case, based on the designs.  There isn't a clear combination of height / width window size classes that fit well here, it's really just portrait vs landscape

feature/auth/src/main/java/org/mozilla/social/feature/auth/chooseServer/ChooseServerViewModel.kt
JohnOberhauser: nope, removed

feature/auth/src/main/java/org/mozilla/social/feature/auth/login/LoginScreen.kt
devotaaabel: IMO @JohnOberhauser what you could do here is something like:
if compact height and width - only show login section
if compact width and medium/extended height - show image and login section vertically
if compact height and medium/extended width - show image and login section horizontally

https://developer.android.com/guide/topics/large-screens/migrate-to-responsive-layouts#things-to-avoid

devotaaabel (approved):
Please still address the last feedback in a follow up

feature/auth/src/main/java/org/mozilla/social/feature/auth/login/LoginScreen.kt
JohnOberhauser: Uhg, it's annoying that multi-window mode can cause the app to be in landscape when height is greater than width

PR #270, opened 2023-11-11
Modules part 4 - cleanup, create mozilla/usecase module, remove network dependency from usecase modules

devotaaabel:


core/repository/mastodon/src/main/java/org/mozilla/social/core/repository/mastodon/AccountRepository.kt
```diff
@@ -106,4 +105,63 @@ class AccountRepository internal constructor(
...
+    @Suppress("MagicNumber")
+    suspend fun updateAccount(
+        displayName: String? = null,
+        bio: String? = null,
+        locked: Boolean? = null,
+        bot: Boolean? = null,
+        avatar: File? = null,
+        header: File? = null,
+        fields: List<Pair<String, String>>? = null
+    ) = accountApi.updateAccount(
```
JohnOberhauser, line 139: I think this needs to return `Account`

devotaaabel: @JohnOberhauser it does, it's called [here](https://github.com/MozillaSocial/mozilla-social-android/pull/270/files#diff-88bb54fe7f92deac76ee93b439a9becba59dec2f50a62e68b5b0febedeba2a96R37) (when you do `fun foo() = bar()` foo returns the same type as bar)

JohnOberhauser (approved):

core/repository/mastodon/src/main/java/org/mozilla/social/core/repository/mastodon/AccountRepository.kt
JohnOberhauser: TIL

PR #279, opened 2023-11-14
Following

JohnOberhauser:
Apologies for the large PR.  I didn't think everything through when I started, so I ended up doing too much in this

- Updating the follower / following UI so that it's one screen with two tabs
  - More UI changes to come in the next PR
- Because we are now making the relationship editable in the new UI, new database tables need to be created
  - Added followers and followings tables + DAOs
  - Added repositories to match the new DAOs
  - Added remote mediators to fetch / update the followers / followings lists

Note: I'm adding the database dependency to the feature module because I need access to the `FollowerWrapper` models to do the mapping in the paging flow.  I'm going to explore what I can do with the paging library to see if I can bring that mapping into the repository layer instead.  That'll happen in my next PR though


core/ui/common/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ dependencies {
     implementation(project(":core:common"))
     implementation(project(":core:repository:mastodon"))
     implementation(project(":core:navigation"))
+    implementation(project(":core:database"))
```
devotaaabel, line 16: I'm wondering if in the meantime it would make sense to make a `pagingrepository` module with the necessary transitive dependencies so we don't have to directly depend on the database module...

core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/remotemediators/FollowersRemoteMediator.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    private var nextKey: String? = null
+
+    @Suppress("ReturnCount")
+    override suspend fun load(
+        loadType: LoadType,
+        state: PagingState<Int, FollowerWrapper>
+    ): MediatorResult {
+        return try {
+            var pageSize: Int = state.config.pageSize
```
devotaaabel, line 38: Did you see the `RefreshXTimeline` use cases I made? I think you could follow a similar pattern here which should make testing the logic a little easier without having to deal with instantiating a remote mediator (also lmk if that naming or even the pattern makes sense or not)

core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/remotemediators/FollowingsRemoteMediator.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.mozilla.social.core.database.model.accountCollections.Followee
+import org.mozilla.social.core.database.model.accountCollections.FolloweeWrapper
+import org.mozilla.social.core.repository.mastodon.AccountRepository
+import org.mozilla.social.core.repository.mastodon.DatabaseDelegate
+import org.mozilla.social.core.repository.mastodon.FollowingsRepository
+import org.mozilla.social.core.repository.mastodon.RelationshipRepository
+import org.mozilla.social.core.usecase.mastodon.account.GetLoggedInUserAccountId
+
+@OptIn(ExperimentalPagingApi::class)
+class FollowingsRemoteMediator internal constructor(
```
devotaaabel, line 19: Same here

feature/followers/src/main/java/org/mozilla/social/feature/followers/FollowersScreen.kt
```diff
@@ -71,94 +81,158 @@ private fun FollowersScreen(
...
+                mutableStateOf(startingTab)
+            }
 
-            val pullRefreshState = rememberPullRefreshState(
-                refreshing = lazyPagingItems.loadState.refresh == LoadState.Loading,
-                onRefresh = { lazyPagingItems.refresh() }
+            MoSoTabRow(
+                selectedTabIndex = selectedTab.ordinal,
+            ) {
+                FollowType.entries.forEach { tabType ->
```
devotaaabel, line 96: Didn't know about `entries`, that's pretty handy. Looks like it was added this year

devotaaabel (approved):
LGTM

core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/remotemediators/FollowersRemoteMediator.kt
JohnOberhauser: I'll take a look at this in a follow up PR

core/ui/common/build.gradle.kts
JohnOberhauser: I'll look into this in the follow up PR

PR #281, opened 2023-11-15
Followers updates part 2

JohnOberhauser:
- Fixing an issue with the tables I created
- Removing the database dependency from the ui:common and feature:follower modules.
- Moving pager creation to the repository

### DIFF
--- a/core/database/src/main/java/org/mozilla/social/core/database/DatabaseMigrations.kt
+++ b/core/database/src/main/java/org/mozilla/social/core/database/DatabaseMigrations.kt
@@ -33,4 +33,19 @@ object DatabaseMigrations {
         columnName = "createdAt"
     )
     class Schema7to8 : AutoMigrationSpec
+
+    @RenameColumn(
+        tableName = "followings",
+        fromColumnName = "followingAccountId",
+        toColumnName = "followeeAccountId",
+    )
+    @DeleteColumn(
+        tableName = "followings",
+        columnName = "relationshipAccountId",
+    )
+    @DeleteColumn(
+        tableName = "followers",
+        columnName = "relationshipAccountId"
+    )
+    class Schema10to11 : AutoMigrationSpec
 }
--- a/core/database/src/main/java/org/mozilla/social/core/database/model/accountCollections/Followee.kt
+++ b/core/database/src/main/java/org/mozilla/social/core/database/model/accountCollections/Followee.kt
@@ -6,17 +6,20 @@ import androidx.room.Relation
 import org.mozilla.social.core.database.model.DatabaseAccount
 import org.mozilla.social.core.database.model.DatabaseRelationship
 
+/**
+ * @param accountId the account ID of the user we are examining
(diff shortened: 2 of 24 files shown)


core/database/src/main/java/org/mozilla/social/core/database/DatabaseMigrations.kt
```diff
@@ -33,4 +33,19 @@ object DatabaseMigrations {
         columnName = "createdAt"
     )
     class Schema7to8 : AutoMigrationSpec
+
+    @RenameColumn(
```
devotaaabel, line 37: I don't think we should be doing database migrations pre-V1. If people run into issues, they can clear their storage

JohnOberhauser: Is there any harm in it though?

devotaaabel: It just feels like unnecessary clutter to me 🤷‍♀️ but it's not a hill I'm gonna die on

JohnOberhauser: We can always delete them all right before we do our real release.  I think it's nice that people using the nightly don't randomly get crashes that cause them to need to reinstall.

core/database/src/main/java/org/mozilla/social/core/database/model/accountCollections/Followee.kt
```diff
@@ -6,31 +6,34 @@ import androidx.room.Relation
...
+/**
+ * @param accountId the account ID of the user we are examining
+ * @param followeeAccountId the account ID of the followee
+ */
 @Entity(
     tableName = "followings",
     primaryKeys = [
         "accountId",
-        "followingAccountId",
+        "followeeAccountId",
```
devotaaabel, line 17: Is there a difference in what you're calling following vs. followee? Can we plz be consistent and just pick one if there's no difference in how you're using them? Or, if `following` is the user-facing copy and we use `followee` internally that's fine too, but it might be helpful to clarify that somewhere

JohnOberhauser: I chose `followee` because it refers to an individual entity, so `following` doesn't make sense, and `followinger` doesn't sound right.  I named the table `followings` because it sounded right, but I could rename it to `followees` I suppose.

devotaaabel: `followinger` 😂 Yea that makes sense. I think it would make sense to rename it `followee` everywhere- honestly "following" just doesn't really make sense at all even though it's the ubiquitous term.... We can think on it though, doesn't have to happen here and now

devotaaabel (approved):
I'm skeptical about us needing the new tables but we can chat and address later.

PR #282, opened 2023-11-15
Content warnings

JohnOberhauser:
- Updating content warning UI in new post screen
- Adding content warnings to post cards


core/designsystem/src/main/java/org/mozilla/social/core/designsystem/icon/MoSoIcons.kt
```diff
@@ -131,6 +131,9 @@ object MoSoIcons {
     @Composable
     fun lock() = painterResource(id = R.drawable.lock)
 
+    @Composable
```
devotaaabel, line 134: This isn't a huge deal but I've been trying to keep these alphabetized (Though it wouldn't be hard to convince me that it's not necessary)

core/ui/postcard/src/main/java/org/mozilla/social/core/ui/postcard/PostCard.kt
```diff
@@ -213,6 +211,116 @@ private fun MetaData(
...
+}
+
+@Composable
+private fun ContentWarning(
+    contentWarningText: String,
+    content: @Composable () -> Unit,
+) {
+    Column {
+        val hasContentWarning by remember(contentWarningText) {
+            mutableStateOf(contentWarningText.isNotBlank())
```
devotaaabel, line 252: I think you want to use `derivedStateOf` instead of `mutableStateOf`

core/ui/postcard/src/main/java/org/mozilla/social/core/ui/postcard/PostCard.kt
```diff
@@ -213,6 +211,116 @@ private fun MetaData(
...
+        }
+        var isShowing by remember(hasContentWarning) {
+            mutableStateOf(!hasContentWarning)
```
devotaaabel, line 255: Same here

devotaaabel (approved):

core/ui/postcard/src/main/java/org/mozilla/social/core/ui/postcard/PostCard.kt
JohnOberhauser: This one has to be mutable since I'm changing the value when the user clicks the content warning

PR #283, opened 2023-11-15
Added about settings

devotaaabel:
I had to make model updates because the existing models were for the v1 API, but it's deprecated so we're using v2.

I'm pulling out a separate ticket for hooking up the links.

### DIFF
--- a/core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/InstanceApi.kt
+++ b/core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/InstanceApi.kt
@@ -1,5 +1,6 @@
 package org.mozilla.social.core.network.mastodon
 
+import kotlinx.serialization.Serializable
 import org.mozilla.social.core.network.mastodon.model.NetworkInstance
 import org.mozilla.social.core.network.mastodon.model.NetworkInstanceRule
 import retrofit2.http.GET
@@ -11,4 +12,10 @@ interface InstanceApi {
 
     @GET("/api/v2/instance")
     suspend fun getInstance(): NetworkInstance
+
+    @GET("/api/v1/instance/extended_description")
+    suspend fun getExtendedDescription(): NetworkExtendedDescription
 }
+
+@Serializable
+data class NetworkExtendedDescription(val content: String)
--- a/core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
+++ b/core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
@@ -21,7 +21,7 @@ fun mastodonNetworkModule(isDebug: Boolean) = module {
             .connectTimeout(OKHTTP_TIMEOUT, TimeUnit.SECONDS)
             .addNetworkInterceptor(HttpLoggingInterceptor().apply {
                 level = if (isDebug) {
-                    HttpLoggingInterceptor.Level.BASIC
+                    HttpLoggingInterceptor.Level.BODY
                 } else {
                     HttpLoggingInterceptor.Level.NONE
                 }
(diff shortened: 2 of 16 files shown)


core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
```diff
@@ -21,7 +21,7 @@ fun mastodonNetworkModule(isDebug: Boolean) = module {
             .connectTimeout(OKHTTP_TIMEOUT, TimeUnit.SECONDS)
             .addNetworkInterceptor(HttpLoggingInterceptor().apply {
                 level = if (isDebug) {
-                    HttpLoggingInterceptor.Level.BASIC
+                    HttpLoggingInterceptor.Level.BODY
```
devotaaabel, line 24: lmk how you feel about this- I think it's convenient to be on by default for debug builds

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/InstanceApi.kt
```diff
@@ -11,4 +12,10 @@ interface InstanceApi {
...
     suspend fun getInstance(): NetworkInstance
-}
\ No newline at end of file
+
+    @GET("/api/v1/instance/extended_description")
+    suspend fun getExtendedDescription(): NetworkExtendedDescription
+}
+
+@Serializable
+data class NetworkExtendedDescription(val content: String)
```
JohnOberhauser, line 21: Can you put this class in it's own file in the models directory?

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
JohnOberhauser, line 24: Body produces a lot of logcat output.  I was hoping we could keep this at basic, and use the network inspector if we need details.  What do you think?

feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutSettingsScreen.kt
```diff
@@ -1,20 +1,146 @@
...
+    Spacer(modifier = Modifier.height(MoSoSpacing.lg))
+    MoSoDivider()
+    Spacer(modifier = Modifier.height(MoSoSpacing.lg))
+}
+
+@Preview
+@Composable
+fun AboutSettingsScreenPreview() {
+    PreviewTheme {
+        MoSoSurface {
```
JohnOberhauser, line 116: FYI, I added this `PreviewTheme` that applies koin, mosoTheme, and mosoSurface, so the `MoSoSurface` here is redundant

JohnOberhauser (approved):

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
JohnOberhauser: Didn't this comment, and left a comment about it below

devotaaabel: I'll prob still change it locally occasionally but that's fine. I'll remove it in the second PR

feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutSettingsScreen.kt
devotaaabel: I'll address this in the next PR

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/InstanceApi.kt
devotaaabel: I at some point want to make the case for putting data classes like these where they are contextually used, but for now I'll add them. Next PR

PR #291, opened 2023-11-17
UI test dependencies

JohnOberhauser:
Setting up ui test dependencies


core/ui/postcard/src/main/java/org/mozilla/social/core/ui/postcard/PostCard.kt
```diff
@@ -395,12 +396,14 @@ private fun BottomRow(
...
     ) {
         BottomIconButton(
+            modifier = Modifier.testTag("reply"),
             onClick = { postCardInteractions.onReplyClicked(post.statusId) },
             painter = MoSoIcons.chatBubbles(),
             count = post.replyCount,
         )
         Spacer(modifier = Modifier.weight(1f))
         BottomIconButton(
+            modifier = Modifier.testTag("boost"),
```
devotaaabel, line 406: I think we should be using [semantics](https://developer.android.com/jetpack/compose/testing#semantics) instead of test tags since we're eventually going to improve the accessibility

build-logic/convention/src/main/kotlin/AndroidLibraryComposeConventionPlugin.kt
```diff
@@ -29,6 +30,7 @@ class AndroidLibraryComposeConventionPlugin : Plugin<Project> {
 
             extensions.configure<LibraryExtension> {
                 configureAndroidCompose(this)
+                configureUiTests(this)
```
devotaaabel, line 33: Sounds like we don't need this line anymore

devotaaabel (approved):

PR #296, opened 2023-11-17
Code cleanup part 1

devotaaabel:
I was originally going to add ktlint rules to our detekt setup, but the `FunctionNaming` rule gave us 2k problems because of the compose convention of using PascalCase instead of camelCase. You can't disable rules in ktlint, so obviously this is a no-go. I think we should look into [slack's compose linting rules](https://slackhq.github.io/compose-lints/) eventually.

So I ended up just running Android Studio's code cleanup, and then ktlint's auto formatting. These are all automated changes.


JohnOberhauser (approved):
RIP git blame :skull:

PR #306, opened 2023-11-22
Favorites part 1

JohnOberhauser:
- Setting up favorites module
- Added empty favorites screen
- Added favorites button in account that navigates to favorites
- Fixed issue with settings showing on other user's profiles


timc-mozilla (approved):

feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
```diff
@@ -310,6 +315,14 @@ private fun OverflowMenu(
                 overflowMenuExpanded.value = false
             },
         ) {
+            if (isUsersProfile) {
```
devotaaabel, line 318: To keep the view dumb (and to support more complicated logic like blocking/unblocking, etc.), I'd prefer the view model emit a state that represents the options that should be displayed. So something like:

```
enum class ProfileDropDownOptions { FAVORITES, BLOCK, UNBLOCK ... }
```
And the view model could emit a list of ProfileDropDownOptions

devotaaabel (approved):

feature/account/src/main/java/org/mozilla/social/feature/account/AccountScreen.kt
JohnOberhauser: noted.  I'm going to hold off on refactoring that in this PR since it affects other parts of the account screen as well

PR #314, opened 2023-12-06
Favorites part 3

JohnOberhauser:
- Setting up favorites view model and UI
- updating undoFavorites use case so it removes the favorite from the favorites timeline


devotaaabel (approved):
Fine to merge, but this seems like a good scoped example of where we could use foreign keys

PR #316, opened 2023-12-08
Forgein Keys

JohnOberhauser:
We need to add foreign keys to the database to prevent any issues when using database relationships

- Removing the hashtag table because we don't use it
- Moving all entity models into a new `entities` directory
- Adding foreign keys to all timelines and the statuses table
- Making the `SaveStatusesToDatabase` use case save all polls and accounts before statuses because of the new foreign key constraints


devotaaabel (approved):
Looks good, ty for doing this. Next time it wouldn't hurt to remove the hashtag timeline table in a separate PR

PR #319, opened 2023-12-13
Setup blocked users list with pagination using room

devotaaabel:
- Setup blocks and mutes daos, remote mediators and repos (mutes list populating as well as button functionality for both lists will be in a separate PR)
- Added `HeaderLink` and `AccountPagingWrapper` which are kind of extensions on top of `MastodonPagingLink`. If there's no objections I can replace existing usages of `MastodonPagingLink` with `HeaderLink` (and maybe rename `HeaderLink` back to `MastodonPagingLink`).

Kind of an open question- I think there's some possibility for code reuse here, but I just didn't want to over-engineer anything.

Sorry this is a big PR


core/database/src/main/java/org/mozilla/social/core/database/model/entities/accountCollections/DatabaseBlock.kt
```diff
@@ -0,0 +1,43 @@
...
+        parentColumns = arrayOf("accountId"),
+        childColumns = arrayOf("accountId"),
+        onDelete = CASCADE,
+        onUpdate = CASCADE,
+    )]
+)
+data class DatabaseBlock(
+    @PrimaryKey val accountId: String,
+    val isBlocked: Boolean,
+    @ColumnInfo(defaultValue = "0")
```
JohnOberhauser, line 26: I don't think you need this annotation.  It was required in some other tables because I had added this field in a new database version after adding the table, so a default value was required for rows that already existed.

JohnOberhauser (approved):

timc-mozilla (approved):

PR #321, opened 2023-12-13
Search part 1

JohnOberhauser:
- Adding a search icon to home
- Setting up search screen with a search bar and tabs
  - Searching doesn't work yet


feature/feed/src/main/java/org/mozilla/social/feed/FeedScreen.kt
```diff
@@ -88,7 +94,21 @@ private fun FeedScreen(
...
+                actions = {
+                    IconButton(
+                        modifier = Modifier
+                            .size(24.dp),
+                        onClick = { feedInteractions.onSearchClicked() }
+                    ) {
+                        Icon(
+                            modifier = Modifier.size(24.dp),
+                            painter = MoSoIcons.magnifyingGlass(),
+                            contentDescription = "",
```
devotaaabel, line 106: I think you want a content description here since there's no text label

devotaaabel (approved):

feature/feed/src/main/java/org/mozilla/social/feed/FeedScreen.kt
JohnOberhauser: Will do this in the next PR

PR #322, opened 2023-12-13
Created DefaultHtmlInteractions and hooked up in about settings screen

devotaaabel:
We could inject this into the `HtmlContent` composable as a default parameter, but it's used a few places so I didn't want to break anything.

### DIFF
--- a/feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutSettingsScreen.kt
+++ b/feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutSettingsScreen.kt
@@ -18,7 +18,6 @@ import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
-import androidx.lifecycle.viewmodel.compose.viewModel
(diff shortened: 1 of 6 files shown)


timc-mozilla (approved):

feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutSettingsScreen.kt
```diff
@@ -125,25 +132,26 @@ fun AboutSettingsScreenPreview() {
...
+                title = "mozilla.social",
+                administeredBy =
+                AccountQuickViewUiState(
+                    accountId = "",
+                    displayName = "Mozilla Social",
+                    webFinger = "@social",
+                    avatarUrl = "",
                 ),
+                contactEmail = "support@mozilla-social.zendesk.com",
+                extendedDescription =
```
JohnOberhauser, line 147: I know you're just formatting here, but we should move this to the strings.xml at some point

JohnOberhauser (approved):

feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutSettingsScreen.kt
devotaaabel: @JohnOberhauser this is in a compose preview

PR #331, opened 2023-12-18
Search part 3

JohnOberhauser:
- Setting up network call
- Setting up database tables

### DIFF
--- a/core/database/src/main/java/org/mozilla/social/core/database/model/entities/hashtagCollections/SearchedHashTag.kt
+++ b/core/database/src/main/java/org/mozilla/social/core/database/model/entities/hashtagCollections/SearchedHashTag.kt
@@ -1,0 +1,37 @@
+package org.mozilla.social.core.database.model.entities.hashtagCollections
+
+import androidx.room.Embedded
+import androidx.room.Entity
+import androidx.room.ForeignKey
+import androidx.room.PrimaryKey
+import androidx.room.Relation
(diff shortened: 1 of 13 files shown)


core/database/src/main/java/org/mozilla/social/core/database/model/entities/hashtagCollections/SearchedHashTag.kt
```diff
@@ -0,0 +1,37 @@
...
+
+data class SearchedHashTagWrapper(
+    @Embedded
+    val searchedHashTag: SearchedHashTag,
+
+    @Relation(
+        parentColumn = "hashTagName",
+        entityColumn = "name",
+    )
+    val status: DatabaseHashTagEntity,
```
devotaaabel, line 36: Is this supposed to be called `status`?

devotaaabel (approved):
Excited for the app to have search 🔥

core/database/src/main/java/org/mozilla/social/core/database/model/entities/hashtagCollections/SearchedHashTag.kt
JohnOberhauser: oops

JohnOberhauser: I'm just going to fix it in the next PR

PR #332, opened 2023-12-19
Updated mutes/blocks lists with clickable accounts

devotaaabel:


JohnOberhauser:
This makes it so that the onclick in the favorites screen doesn't get called.  It still navigates, but if we need some analytics to happen on a particular screen, this will override it.  I think it would be better to just have each screen handle the clicks themselves and make this UI component as dumb as possible.  That would also allow you to remove the navigation dependency from this module.

timc-mozilla (approved):

devotaaabel:
@JohnOberhauser I took the navigation out of the AccountQuickView

JohnOberhauser (approved):

PR #335, opened 2023-12-20
Search part 5

JohnOberhauser:
- Getting Top tab set up and working

### DIFF
--- a/core/database/src/main/java/org/mozilla/social/core/database/dao/SearchDao.kt
+++ b/core/database/src/main/java/org/mozilla/social/core/database/dao/SearchDao.kt
@@ -37,9 +37,19 @@ interface SearchDao {
     fun hashTagsPagingSource(): PagingSource<Int, SearchedHashTagWrapper>
 
     @Query(
-        "DELETE FROM searchedAccounts ",
+        "DELETE FROM searchedAccounts "
     )
-    suspend fun deleteAllSearches()
+    suspend fun deleteAllSearchedAccounts()
+
+    @Query(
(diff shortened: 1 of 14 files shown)


core/database/src/main/java/org/mozilla/social/core/database/dao/SearchDao.kt
devotaaabel, line 1: Should there be a separate dao for each table?

JohnOberhauser: I originally had this single dao because I was going to have queries that referenced all three tables, but I didn't end up doing that.  I'm still okay keeping a single dao because of how closely related the tables are.  And the API is all one call / response so it kind of fits to have a single dao as well.

devotaaabel (approved):

feature/search/src/main/java/org/mozilla/social/search/SearchScreen.kt
```diff
@@ -94,27 +123,42 @@ private fun SearchScreen(
...
+                        uiState = uiState,
+                        searchInteractions = searchInteractions,
+                    )
+                    ListContent(
+                        uiState = uiState,
+                        searchInteractions = searchInteractions,
+                        postCardInteractions = postCardInteractions,
+                    )
+                }
+                androidx.compose.animation.AnimatedVisibility(
```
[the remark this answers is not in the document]
JohnOberhauser: For some reason it thinks I'm trying to use `ColumnScope.AnimatedVisibility` when it's not fully qualified

PR #339, opened 2023-12-22
Search part 6 

JohnOberhauser:
- Hooking up account tab
- Adding an empty state to the top tab

### DIFF
--- a/core/ui/common/src/main/java/org/mozilla/social/core/ui/common/paging/SearchPagingColumn.kt
+++ b/core/ui/common/src/main/java/org/mozilla/social/core/ui/common/paging/SearchPagingColumn.kt
@@ -1,0 +1,112 @@
+package org.mozilla.social.core.ui.common.paging
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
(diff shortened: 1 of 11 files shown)


core/ui/common/src/main/java/org/mozilla/social/core/ui/common/paging/SearchPagingColumn.kt
```diff
@@ -0,0 +1,112 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import androidx.paging.LoadState
+import androidx.paging.compose.LazyPagingItems
+import org.mozilla.social.core.designsystem.theme.MoSoTheme
+import org.mozilla.social.core.ui.common.animation.DelayedVisibility
+import org.mozilla.social.core.ui.common.error.GenericError
+
+@Composable
+fun <A : Any> SearchPagingColumn(
```
devotaaabel, line 26: Is this only for search?

devotaaabel (approved):

core/ui/common/src/main/java/org/mozilla/social/core/ui/common/paging/SearchPagingColumn.kt
JohnOberhauser: It could be used for other things technically.  But I think the only possible usage at the moment is search

JohnOberhauser: I might rename this

PR #341, opened 2023-12-22
Search part 7 - Statuses

JohnOberhauser:
- Getting the statuses tab to work
- Fixing an issue where the "in reply to" information was not being fetched for the top tab

### DIFF
--- a/core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/status/GetInReplyToAccountNames.kt
+++ b/core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/status/GetInReplyToAccountNames.kt
@@ -1,0 +1,28 @@
+package org.mozilla.social.core.usecase.mastodon.status
+
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import org.mozilla.social.core.model.Status
+import org.mozilla.social.core.repository.mastodon.AccountRepository
+
+class GetInReplyToAccountNames internal constructor(
+    private val accountRepository: AccountRepository,
+) {
+
+    suspend operator fun invoke(
+        statuses: List<Status>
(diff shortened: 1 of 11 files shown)


core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/status/GetInReplyToAccountNames.kt
```diff
@@ -0,0 +1,28 @@
...
+    private val accountRepository: AccountRepository,
+) {
+
+    suspend operator fun invoke(
+        statuses: List<Status>
+    ): List<Status> =
+        coroutineScope {
+            statuses.map { status ->
+                async {
+                    status.copy(
```
devotaaabel, line 18: Is this because of a mastodon bug? If so we should probably talk to the platform team about it. If the original call isn't coming back with all the info, I feel like we should use different models for with/without

devotaaabel (approved):

core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/status/GetInReplyToAccountNames.kt
JohnOberhauser: Yeah, it's a mastodon issue.  I'm not sure if it's a bug or if it was designed this way for a reason.  

I think keeping it as one model makes things easier.  Otherwise we would need a separate database table for statuses that are replies.  That would also make all of our timeline tables a lot more complicated.

JohnOberhauser: If you want to continue this conv, you can message me on slack or add this to the TIP agenda.  I'm going to merge this PR

PR #343, opened 2023-12-22
Search part 8 - hash tags

JohnOberhauser:
- Showing hash tags in the hash tags tab
- Adding use cases for following and unfollowing hash tags


core/ui/common/src/main/java/org/mozilla/social/core/ui/common/hashtag/quickview/HashTagQuickView.kt
```diff
@@ -0,0 +1,102 @@
...
+                    .align(Alignment.Center),
+                painter = MoSoIcons.hash(),
+                contentDescription = "",
+            )
+        }
+
+
+        Spacer(modifier = Modifier.width(MoSoSpacing.sm))
+
+        Column {
```
devotaaabel, line 56: Is this column serving a purpose?

devotaaabel (approved):
Looks good, though I want to reiterate that we really need to spend some time finding ways to reuse paging code using composition over inheritance

core/ui/common/src/main/java/org/mozilla/social/core/ui/common/hashtag/quickview/HashTagQuickView.kt
JohnOberhauser: Nope, I deleted it

JohnOberhauser:
> Looks good, though I want to reiterate that we really need to spend some time finding ways to reuse paging code using composition over inheritance

I'll add something about this in the TIP agenda.

PR #344, opened 2023-12-24
Fetching in reply to account name issue

JohnOberhauser:
I came across a post that has an "in reply to" account ID that does not correspond to any account, so we get a 404 when trying to fetch it.  Adding some code into the use case that checks for 404 and assumes the post is not a reply if there is one.

The post with the issue: https://mozilla.social/@lsamiee/111597399441121379

- Also making sure all remote mediators use this use case instead of the old extension function


devotaaabel (changes requested):
I would honestly really prefer that for expected and predictable situations like these, we wrap the result in a sealed class (something like `sealed class ApiResponse { data class Success() ... data class Error()}`). The difference in my mind is that a sealed class enforces that the call site has to handle the error (or at least be aware that it's a possibility), vs. exceptions which can slip through the cracks (and IMO are for situations where fundamental assumptions are broken). But we can talk about it when you're back

JohnOberhauser:
> I would honestly really prefer that for expected and predictable situations like these, we wrap the result in a sealed class (something like `sealed class ApiResponse { data class Success() ... data class Error()}`). The difference in my mind is that a sealed class enforces that the call site has to handle the error (or at least be aware that it's a possibility), vs. exceptions which can slip through the cracks (and IMO are for situations where fundamental assumptions are broken). But we can talk about it when you're back

hmm, okay, let's chat on slack / zoom

devotaaabel (approved):

PR #346, opened 2024-01-03
Follow hashtag button

JohnOberhauser:
- Adding a follow button to the hashtag screen
- Because we need to get the follow state from network, I'm also adding loading and error states to the hashtag screen


core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/MastodonUsecaseModule.kt
```diff
@@ -209,4 +210,10 @@ val mastodonUsecaseModule =
         singleOf(::GetDomain)
         singleOf(::SearchAll)
         singleOf(::GetInReplyToAccountNames)
+
+        single {
+            GetHashTag(
+                hashtagRepository = get(),
+            )
+        }
```
devotaaabel, line 218:
```suggestion
        singleOf(::GetHashTag)
```

devotaaabel (approved):

core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/MastodonUsecaseModule.kt
JohnOberhauser: `GetHashTag` has another parameter for the coroutine dispatcher with a default value, so we have to use the normal constructor pattern

Issue #348, opened 2024-01-04
sign in with friendica account not possible


foss-:
MoSo Nightly 1.0.27 Pre-release
Android 13.1
This is Friendica, version 2024.03-dev that is running at the web location https://pirati.ca. The database version is 1545/1545, the post update version is 1544/1544.

**Reproduce**

1. Choose a server
2. enter pirati.ca
3. webUI opens asking to `Authorize application connection`, confirm with `Yes`

**Currrently**

back a `Choose a server` screen, go in a loop

**Expected**

Accept login and authorization.

JohnOberhauser:
Thanks for reporting this.  I've opened a PR to fix the issue https://github.com/MozillaSocial/mozilla-social-android/pull/349

foss-:
Thanks for tackling this so swiftly. Sign-in works as expected in Nightly 0.1.0.30.

PR #349, opened 2024-01-05
Removing broken field from account

JohnOberhauser:
This change will fix this issue https://github.com/MozillaSocial/mozilla-social-android/issues/348

The mastodon API spec says the `last_status_at` field in an `Account` body should be in a format like this `2019-11-24T15:49:42.251Z`, however it's coming back like this `2019-11-24`.  We were parsing it the way mastodon gives it to us, but it's incorrect according to the docs.  Friendica sends back the correct format, which is why Friendica sign in wasn't working.  I'm just removing the field because it's broken in mastodon and we aren't using it anyways.


devotaaabel (approved):
Nice!

PR #353, opened 2024-01-05
Periodically purge database

JohnOberhauser:
TODO: test purging when activity is open


app/src/main/AndroidManifest.xml
```diff
@@ -33,5 +33,17 @@
                     android:scheme="mozsoc"/>
             </intent-filter>
         </activity>
+
+        <!--> Remove the work manager provider so we can use koin's https://insert-koin.io/docs/reference/koin-android/workmanager/ <-->
```
devotaaabel, line 37: Is this a todo?

core/workmanager/src/main/java/org/mozilla/social/core/workmanager/DatabasePurgeWorker.kt
```diff
@@ -0,0 +1,190 @@
...
+                searchRepository.deleteSearchResults()
+                timelineRepository.deleteLocalTimeline()
+                timelineRepository.deleteFederatedTimeline()
+                timelineRepository.deleteAllHashTagTimelines()
+                timelineRepository.deleteAllAccountTimelines()
+
+                hashtagRepository.deleteAll()
+                relationshipRepository.deleteAll()
+
+                val homeStatuses = timelineRepository.getTopHomePostsFromDatabase(40)
```
devotaaabel, line 79: We need to be careful of the complexity here. I imagine it getting more complex when we implement timeline markers.

What do you think about including a last accessed timestamp to all database rows, and just deleting everything older than a certain age (maybe 20 days)? Certain rows that never should be deleted (like a users own account) could have a special max timestamp

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsInteractions.kt
```diff
@@ -2,4 +2,9 @@ package org.mozilla.social.feature.settings
 
 interface SettingsInteractions {
     fun onScreenViewed() = Unit
+    fun onAboutClicked() = Unit
```
devotaaabel, line 5: Same comment as last PR about default values

devotaaabel (commented):
Thanks a bunch for setting all this up 🙏 Would you mind doing a decision doc for this, so we can just make sure we're considering all important factors/getting input before such a big change?

devotaaabel:
Also- I understand it's probably too late, but for the future, it would be nice if the developer options was in a separate PR

app/src/main/AndroidManifest.xml
JohnOberhauser: No, this is a note on why it's there

core/workmanager/src/main/java/org/mozilla/social/core/workmanager/DatabasePurgeWorker.kt
JohnOberhauser: I want to address this on a zoom call

feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsInteractions.kt
JohnOberhauser: I just realized this interface is being used in places I didn't expect.  I'd like to fix this in a separate PR.  I'll create a ticket for it and work on it once this is merged

core/workmanager/src/main/java/org/mozilla/social/core/workmanager/DatabasePurgeWorker.kt
```diff
@@ -0,0 +1,167 @@
...
+    private val statusRepository: StatusRepository,
+    private val timelineRepository: TimelineRepository,
+): CoroutineWorker(
+    context,
+    workerParams,
+) {
+
+    private val loggedInUserId = getLoggedInUserAccountId()
+
+    override suspend fun doWork(): Result {
```
devotaaabel, line 54: I like the idea here, but just a note that `Result` is meant for [general purpose APIs and not domain-specific error conditions](https://github.com/Kotlin/KEEP/blob/master/proposals/stdlib/result.md#error-handling-style-and-exceptions). I think we should use a custom sealed class here in place of Result.

[Relevant SO discussion about using Kotlin.Result](https://stackoverflow.com/a/70847760)

core/workmanager/src/main/java/org/mozilla/social/core/workmanager/DatabasePurgeWorker.kt
```diff
@@ -0,0 +1,167 @@
...
+         * It then observes that work and restarts the app when complete.
+         * We need to restart the app if it is open because purging the database could put us in
+         * a weird state.
+         * The user won't be using the app during this because there is a constraint put on the work request.
+         */
+        fun setupPurgeWork(
+            activity: Activity,
+            lifecycleCoroutineScope: LifecycleCoroutineScope,
+        ) {
+            val workRequest = PeriodicWorkRequestBuilder<DatabasePurgeWorker>(7, TimeUnit.DAYS)
```
devotaaabel, line 98: Are we sure we want to do this once a week? I'm still wondering if we could vacuum once a week and purge other stuff once a month or less. Or, is the thinking that we can do it more often because we're only deleting old statuses?

core/workmanager/src/main/java/org/mozilla/social/core/workmanager/DatabasePurgeWorker.kt
JohnOberhauser: The `doWork` function is overridden from `CoroutineWorker` which is part of work manager.  I can't change the return type.

core/workmanager/src/main/java/org/mozilla/social/core/workmanager/DatabasePurgeWorker.kt
JohnOberhauser: I think 7 days is okay since we are only deleting old statuses.  We can always change it later if we have reason to.

devotaaabel (approved):

PR #354, opened 2024-01-07
About screen update

JohnOberhauser:
- Adding loading and error states to the about screen
- This fixes a crash when there is an error loading data


feature/settings/src/main/java/org/mozilla/social/feature/settings/about/AboutInteractions.kt
```diff
@@ -0,0 +1,6 @@
+package org.mozilla.social.feature.settings.about
+
+interface AboutInteractions {
+    fun onOpenSourceLicensesClicked() = Unit
```
devotaaabel, line 4: The default values here are risky and don't really serve a purpose, because we always want both of those functions implemented

devotaaabel (approved):

PR #358, opened 2024-01-08
Updated PR workflow to run UI tests on a macos runner

devotaaabel:


JohnOberhauser:
Are there any limits when using a macos runner?  I think those require apple hardware so they are typically more expensive / limited

.github/workflows/pr.yml
```diff
@@ -22,6 +22,12 @@ jobs:
       env:
         GRADLE_OPTS: -Dorg.gradle.daemon=false
       run: ci/unsigned-release-build.sh
+    - name: Archive the APK
```
JohnOberhauser, line 25: This will archive the apk for every change on a PR.  I think we need a new workflow that happens on pushes to main to do this instead

.github/workflows/nightly.yml
```diff
@@ -1,8 +1,9 @@
 name: nightly
 
+# Run every weekday at 12:45am
 on:
-  push:
-    branches: [ "main" ]
+  schedule:
+    - cron: 45 0 * * 1-5
```
JohnOberhauser, line 6: Can we make this include weekends in case we merge stuff over the weekend?

JohnOberhauser (approved):

PR #359, opened 2024-01-09
Adding developer options screen

JohnOberhauser:
I pulled the changes out of https://github.com/MozillaSocial/mozilla-social-android/pull/353

- Adding a developer options screen to settings that only shows in debug mode


feature/settings/src/main/java/org/mozilla/social/feature/settings/SettingsInteractions.kt
```diff
@@ -2,4 +2,9 @@ package org.mozilla.social.feature.settings
 
 interface SettingsInteractions {
     fun onScreenViewed() = Unit
```
devotaaabel, line 4: Don't forget to remove the default values plz (in next PR is fine)

devotaaabel (approved):

PR #360, opened 2024-01-09
Adding sematics header to splash screen (accessibility)

audie1:
Adding header semantics to title composable in splash screen.


audie1:
Holding this work for now, will revisit after deeming necessity.

PR #361, opened 2024-01-09
Adding bottom nav content descriptions for accessibility

audie1:
Added semantics to bottom nav composables for the icons, the screen reader is now pronouncing the name of each bottom nav option.


devotaaabel (approved):
Nice, thanks for doing this!

PR #363, opened 2024-01-10
Changing beta label to alpha

JohnOberhauser:
- Changing beta label to alpha


feature/auth/src/main/res/values/strings.xml
```diff
@@ -3,7 +3,7 @@
     <string name="login_button">Login</string>
     <string name="title_text">mozilla.social</string>
 
-    <string name="beta_badge">Beta</string>
+    <string name="beta_badge">Alpha</string>
```
devotaaabel, line 6: Might as well rename `beta_badge` to `alpha_badge`

devotaaabel (approved):

PR #372, opened 2024-01-16
Accessibility improvements for postcards

audie1:
Setting avatar image content description


devotaaabel (approved):
Looks good, though I don't think you're merging the semantic tree in this PR

audie1:
Thanks for the reviews @JohnOberhauser and @devotaaabel ... 

I'm going to convert this to a draft for now because I can't get the desired improvements by merging the semantic tree for the postcard composable in PostCard.kt but I'm still tinkering with things. The fact that the postcard is built partially from the HTML of the body text and that's not being merged with my initial attempts seems to be an issue. I'll keep trying a few things and then might need to phone a friend for a resolution if I can't figure it out.

devotaaabel:
@audie1 feel free to remove that bullet point from the description and merge this

audie1:
Ok I'll give it one more try and then I'll remove the bullet point if no luck @devotaaabel

audie1:
Merging this PR, and removing the semantic tree merging point for now, going to revisit that later.

PR #382, opened 2024-01-18
Integrate sentry sdk to enable crash reporting

devotaaabel:
- Setup Sentry with gradle plugin and some default values
- Setup "environment" option to refer to the build type
- Added `unsignedRelease` build type
- Added `sentry.properties` to files to encrypt/decrypt
- Encrypted `sentryDsn` value in `secret.properties` (@MozillaSocial/android you'll need to run `./decrypt` locally to be able to send sentry events)
- Setup Timber to log to sentry on anything other than a debug build using the [sentry-android-timber](https://docs.sentry.io/platforms/android/integrations/timber/) library


app/build.gradle.kts
```diff
@@ -1,6 +1,9 @@
 plugins {
     id("org.mozilla.social.android.application")
     id("org.mozilla.social.android.application.compose")
+    id("org.mozilla.social.android.application.secrets")
```
JohnOberhauser, line 4: Where is this coming from?  It's not defined in our build-logic conventions.

app/build.gradle.kts
```diff
@@ -1,6 +1,9 @@
...
+
+    id("io.sentry.android.gradle") version "4.2.0"
```
JohnOberhauser, line 6: Can you add this to the `libs.versions.toml` file and apply it here using `alias(sentry)`?

app/build.gradle.kts
devotaaabel: https://github.com/MozillaSocial/mozilla-social-android/blob/main/build-logic/convention/src/main/kotlin/org/mozilla/social/android.application.secrets.gradle.kts

JohnOberhauser (approved):

PR #383, opened 2024-01-18
Notification account relationship setup

JohnOberhauser:
- Adding some boilerplate stuff to include a user's relationship for the follow and follow request notifications
  - Fetching relationships for these types
  - Saving them to the database
- Database migration


core/database/src/main/java/org/mozilla/social/core/database/model/wrappers/NotificationWrapper.kt
```diff
@@ -46,6 +47,12 @@ data class NotificationWrapper(
         entityColumn = "pollId",
     )
     val boostedPoll: DatabasePoll?,
+
+    @Relation(
+        parentColumn = "relationshipAccountId",
+        entityColumn = "accountId",
+    )
+    val relationship: DatabaseRelationship?,
```
devotaaabel, line 55: Why are we adding this in the db? I thought we were trying not to diverge from the way objects are modeled in the API. Couldn't this be fetched/combined when mapping the paging data in a use case?

core/repository/mastodon/src/main/java/org/mozilla/social/core/repository/mastodon/model/notifications/ExternalToDatabase.kt
devotaaabel, line 1: Why are the status fields added separately to the DatabaseNotification instead of just a reference to the status like in the API model?

core/repository/mastodon/src/main/java/org/mozilla/social/core/repository/mastodon/NotificationsRepository.kt
```diff
@@ -15,14 +16,17 @@ import org.mozilla.social.core.database.model.entities.notificationCollections.M
...
 import org.mozilla.social.core.repository.mastodon.model.notifications.toDatabase
 import org.mozilla.social.core.repository.mastodon.model.notifications.toExternal
 import retrofit2.HttpException
 
 class NotificationsRepository(
-    val api: NotificationsApi,
-    val dao: NotificationsDao,
+    private val api: NotificationsApi,
+    private val dao: NotificationsDao,
+    private val accountApi: AccountApi,
```
devotaaabel, line 29: I thought we said max of one dao and one api per repository?

devotaaabel (commented):
It would be helpful if you could explain your thinking behind diverging from our repository and database layers mirroring the API layer

core/database/src/main/java/org/mozilla/social/core/database/model/wrappers/NotificationWrapper.kt
JohnOberhauser: This is in a wrapper, so it's not part of the notification entity / table itself.  We do this for all of our collections, for example `BlockWrapper` here https://github.com/MozillaSocial/mozilla-social-android/blob/main/core/database/src/main/java/org/mozilla/social/core/database/model/entities/accountCollections/DatabaseBlock.kt

core/repository/mastodon/src/main/java/org/mozilla/social/core/repository/mastodon/model/notifications/ExternalToDatabase.kt
JohnOberhauser: `DatabaseNotification` is the object that defines the `notifications` table.  The statuses themselves are saved in the `statuses` table.  They are not embedded in the `notifications` table.  We only need the ID's associated with the statuses.  Then we can use the `@Relationship` functionality in room to rebuild the notification when accessing it.

core/repository/mastodon/src/main/java/org/mozilla/social/core/repository/mastodon/NotificationsRepository.kt
devotaaabel: > I'm making the extra network call here in the repository instead of in a specific use case because I want the domain Notification.Follow model to embed the Relationship model, but our use case layer doesn't (and shouldn't) have access to the network models. So I need to have the relationship models at the time of mapping from the network model to the domain model.

Can you elaborate on this? Why do you need the network model? Why can't you combine the domain models of follow/relationship in the use case layer?

JohnOberhauser:
After looking into how notifications work server side, we don't need this PR anymore

PR #393, opened 2024-01-22
Updating posting polls UI

JohnOberhauser:
- Updating UI for posting polls

Screenshot_20240122_173526


devotaaabel:
I don't think the "Hide results until complete" checkbox should be on the same horizontal line as the pickers, even if it scrolls it looks bad (and can be hard to see it's even an option if the phone width cuts it off completely)

JohnOberhauser:
> I don't think the "Hide results until complete" checkbox should be on the same horizontal line as the pickers, even if it scrolls it looks bad (and can be hard to see it's even an option if the phone width cuts it off completely)

I'm just going off the designs

timc-mozilla (approved):

PR #394, opened 2024-01-22
Adding a content description to the Search button for accessibility

audie1:
Adding content desc to Search button in strings.xml


audie1:
Ready for review :)

feature/feed/src/main/java/org/mozilla/social/feed/FeedScreen.kt
```diff
@@ -105,7 +108,7 @@ private fun FeedScreen(
                         Icon(
                             modifier = Modifier.size(24.dp),
                             painter = MoSoIcons.magnifyingGlass(),
-                            contentDescription = "",
+                            contentDescription = searchButton,
```
devotaaabel, line 111: You can just inline it (so this should be `contentDescription = stringResource(R.string.search_button)`)

audie1: Thanks @devotaaabel I made these changes.

JohnOberhauser (approved):

PR #396, opened 2024-01-23
About libraries cleanup

JohnOberhauser:
- Changing `ImmutableList`s to `List`
- removing hard coded strings


devotaaabel (approved):
ImmutableList was added as a dependency for this so I think you can remove that too

PR #406, opened 2024-01-25
basic feature for edit post

timc-mozilla:
Edit Post basic features contain text and content warning
does not actually update post, posts another post with the different content. 
does not show media.


core/navigation/src/main/java/org/mozilla/social/core/navigation/NavigationDestination.kt
```diff
@@ -112,27 +112,27 @@ sealed class NavigationDestination(
         }
     }
 
-    data class NewPost(val replyStatusId: String? = null) : NavigationDestination(
+    data class NewPost(val replyStatusId: String? = null, val isEditingId: String? = null) : NavigationDestination(
```
JohnOberhauser, line 115: Can you rename `isEditingId` to something like `editStatusId`.  `is` is typically used only in booleans.

JohnOberhauser (approved):

PR #411, opened 2024-01-26
Moving search button

JohnOberhauser:
- Removing search from feed screen
- Fixing mozilla logo on feed screen
- Changing explore / discover icon to the search icon
- Adding the search bar to the explore tab


core/designsystem/src/main/res/drawable/mozilla_logo_inverse.xml
devotaaabel, line 1: Can't you just set the tint/background tint so we don't need two assets?

feature/feed/src/main/java/org/mozilla/social/feed/FeedScreen.kt
```diff
@@ -81,96 +76,117 @@ private fun FeedScreen(
...
+                            .background(
+                                if (isSystemInDarkTheme()) {
+                                    FirefoxColor.White
+                                } else {
+                                    FirefoxColor.Black
+                                }
+                            ),
                         painter = mozillaLogo(),
                         contentDescription = "mozilla logo",
+                        tint = if (isSystemInDarkTheme()) {
```
devotaaabel, line 99: This color should live in the theme, maybe we need a new value for logoColor

feature/feed/src/main/java/org/mozilla/social/feed/FeedScreen.kt
```diff
@@ -81,96 +76,117 @@ private fun FeedScreen(
...
-                    TimelineType.FEDERATED -> federatedScrollState
+    MoSoTabRow(
+        selectedTabIndex = selectedTimelineType.ordinal,
+        divider = {},
+    ) {
+        TimelineType.entries.forEach { timelineType ->
+            MoSoTab(
+                modifier =
+                Modifier
+                    .height(40.dp),
```
devotaaabel, line 144: Where did 40 come from? I tend to prefer hardcoding padding values instead of height values to be more dynamic based on the content/text size, I wonder if that's possible here

feature/search/src/main/java/org/mozilla/social/search/SearchScreen.kt
```diff
@@ -111,46 +115,47 @@ private fun SearchScreen(
...
+                                searchInteractions.onQueryTextChanged("")
+                                searchFocusRequester.requestFocus()
                             }
-                        },
-                    )
-                }
-            )
+                        ) {
+                            Icon(
+                                modifier = Modifier.size(16.dp),
```
devotaaabel, line 149: Use `MoSoIcons.Sizes.small` here

feature/discover/src/main/java/org/mozilla/social/feature/discover/DiscoverScreen.kt
```diff
@@ -68,16 +73,41 @@ private fun DiscoverScreen(
...
-        ) {
+        Column {
+            Box(
+                modifier = Modifier
+                    .padding(
+                        start = MoSoSpacing.md,
+                        end = MoSoSpacing.md,
+                        // 14 on top should make the search bar align perfectly with the
+                        // search bar on the search screen
+                        top = 14.dp,
```
devotaaabel, line 88: Maybe this would be useful as a val in an object in the `MoSoSearchBar` file

devotaaabel (approved):

feature/feed/src/main/java/org/mozilla/social/feed/FeedScreen.kt
JohnOberhauser: Not sure, I just copy pasted what was already there.  Just moving the code into it's own composable, not changing it in this PR

feature/discover/src/main/java/org/mozilla/social/feature/discover/DiscoverScreen.kt
JohnOberhauser: hmm, I don't think it belongs there.  It's not `MoSoSearchBar`'s responsibility to handle this type of value.  I'm not sure there is a good place for a constant for this right now.  I'm just going to leave it where it is with the comment.

PR #413, opened 2024-01-28
Sentry update

JohnOberhauser:
- I'm disabling auto init for sentry and moving the sentry settings out of the manifest because I like the keep the manifest as lean as possible
- Disabling screenshots for sentry
- Reducing sentry diagnostic level
  - While I was thinking about reducing logcat output, I set the http output to BASIC.  @devotaaabel let me know if that's okay


app/src/main/java/org/mozilla/social/utils/SentryManager.kt
```diff
@@ -0,0 +1,34 @@
...
+                dsn = BuildConfig.sentryDsn
+                isDebug = BuildConfig.DEBUG
+                environment = BuildConfig.BUILD_TYPE
+                isEnableUserInteractionTracing = true
+                isAttachScreenshot = false
+                isAttachViewHierarchy = true
+                sampleRate = 1.0
+                profilesSampleRate = 1.0
+                addIntegration(
+                    SentryTimberIntegration(
```
JohnOberhauser, line 26: this wasn't added in debug builds before.  Does it matter if it's added to debug builds?

devotaaabel: That was on purpose. Were you thinking we should turn it on for debug builds? My thinking was that we have log output for debug builds, and we're the only people on debug builds anyway (also it's a recommended configuration and I thought it made sense)

JohnOberhauser: Okay, I added the check back in

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
```diff
@@ -0,0 +1,36 @@
+package org.mozilla.social.utils
+
+import android.content.Context
+import io.sentry.SentryLevel
+import io.sentry.android.core.SentryAndroid
+import io.sentry.android.timber.SentryTimberIntegration
+import org.mozilla.social.BuildConfig
+
+object SentryManager {
```
devotaaabel, line 9: Can you make this an injectable class please

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
```diff
@@ -0,0 +1,36 @@
...
+                environment = BuildConfig.BUILD_TYPE
+                isEnableUserInteractionTracing = true
+                isAttachScreenshot = false
+                isAttachViewHierarchy = true
+                sampleRate = 1.0
+                profilesSampleRate = 1.0
+                if (!BuildConfig.DEBUG) {
+                    addIntegration(
+                        SentryTimberIntegration(
+                            minEventLevel = SentryLevel.ERROR,
```
devotaaabel, line 28: This is initializing timber for release builds. I had timber/sentry in the same function so that was clear

app/src/main/java/org/mozilla/social/MainApplication.kt
```diff
@@ -44,24 +45,16 @@ class MainApplication : Application(), ImageLoaderFactory {
...
-                    SentryTimberIntegration(
-                        minEventLevel = SentryLevel.ERROR,
-                        minBreadcrumbLevel = SentryLevel.INFO
-                    )
-                )
-            } else {
-                Timber.plant(Timber.DebugTree())
-            }
+    private fun initializeTimber() {
+        if (BuildConfig.DEBUG) {
```
devotaaabel, line 56: If you don't move this back to a combined function, maybe a comment here could be useful, just mentioning that timber is initialized for release builds in the SentryManager

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
```diff
@@ -24,7 +24,7 @@ val mastodonNetworkModule =
                     HttpLoggingInterceptor().apply {
                         level =
                             if (BuildConfig.DEBUG) {
-                                HttpLoggingInterceptor.Level.BODY
+                                HttpLoggingInterceptor.Level.BASIC
```
devotaaabel, line 27: Why did you switch this to basic? I prefer using filters to reduce stuff like this if it's too much in logcat, but I think it's nice having it default to being pretty verbose so we don't have to hunt down where this is initialized to update the log level if we want to see more info (though I'm open to other opinions)

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
```diff
@@ -0,0 +1,36 @@
...
+
+object SentryManager {
+
+    fun initialize(
+        context: Context,
+    ) {
+        SentryAndroid.init(context) { options ->
+            options.apply {
+                setDiagnosticLevel(SentryLevel.ERROR)
+                dsn = BuildConfig.sentryDsn
```
devotaaabel, line 17: You can remove the manifest placeholders from [this PR](https://github.com/MozillaSocial/mozilla-social-android/pull/382/files) since you moved these from the manifest

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
JohnOberhauser: If I'm looking for anything more than basic information, I find it easier to use the network inspector.  But if you prefer to use logcat for viewing network details, I can keep it as `BODY`.

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
JohnOberhauser: Hmm, I'd rather not, because I want sentry to be initialized before Koin, in case some other class that is initialized with koin (or koin itself) causes a crash.

core/network/mastodon/src/main/java/org/mozilla/social/core/network/mastodon/MastodonNetworkModule.kt
devotaaabel: That's fair, though you have to already have the network inspector running. I'm fine with updating it to basic to see how it goes but if it's annoying I might update it back to body

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
JohnOberhauser: I'm just moving the initialization back to the main application class

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
JohnOberhauser: I'm moving the initialization back to the main application class

app/src/main/java/org/mozilla/social/MainApplication.kt
JohnOberhauser: I moved it back

app/src/main/java/org/mozilla/social/utils/SentryManager.kt
devotaaabel: I guess in that case I don't see why it's in a separate manager, when other things that are initialized in the application don't. But no biggie

app/src/main/java/org/mozilla/social/MainApplication.kt
```diff
@@ -44,25 +44,38 @@ class MainApplication : Application(), ImageLoaderFactory {
...
+                        )
                     )
-                )
-            } else {
-                Timber.plant(Timber.DebugTree())
+                }
             }
         }
+        // Timber is initialized in release builds via the Timber / Sentry integrations above ^
+        if (BuildConfig.DEBUG) {
```
devotaaabel, line 76: nit- having `if(debug)` and `if(!debug)` in different levels of the hierarchy feels a little more convoluted than the `if/else` that was there before

devotaaabel (approved):

PR #414, opened 2024-01-28
Fixing login crash

JohnOberhauser:
Fixing a crash when a user adds `http://` or `https://` to their server's domain


core/usecase/mastodon/src/main/java/org/mozilla/social/core/usecase/mastodon/auth/Login.kt
```diff
@@ -104,6 +106,17 @@ class Login(
         clientSecret = null
     }
 
+    private fun extractHost(domain: String): String {
+        return buildString {
```
devotaaabel, line 110: Can you use `Uri.parse(domain).host` here instead?

JohnOberhauser: That will return null if there is no `http://` or `https://`.  That's why I need to do the `buildString` to make sure the url starts with `https://`

devotaaabel: Can you try `Uri.fromParts`? I'm not 100% sure as I haven't used it before, but it looks like what you need

JohnOberhauser: Hmm, I tried but couldn't figure out a way to do it.

devotaaabel (approved):

PR #415, opened 2024-01-28
New post hashtag and account issues

JohnOberhauser:
There were bugs with highlighting, searching, and replacing hashtags and accounts in the new post screen.  The bugs stemmed from the algorithm not taking into consideration new line characters.

- fixed issues with highlighting, searching, and replacing hashtags and accounts
- added some extra unit tests

### DIFF
--- a/core/common/src/main/java/org/mozilla/social/common/utils/TextFieldValueExtensions.kt
+++ b/core/common/src/main/java/org/mozilla/social/common/utils/TextFieldValueExtensions.kt
@@ -6,32 +6,37 @@ import androidx.compose.ui.text.TextRange
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.input.TextFieldValue
 import androidx.compose.ui.text.withStyle
+import kotlin.math.min
 
 /**
  * @return the account string a user is typing if they are typing one, otherwise null
  */
-fun TextFieldValue.accountText(): String? = findSymbolText('@')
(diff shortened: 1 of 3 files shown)


core/common/src/main/java/org/mozilla/social/common/utils/TextFieldValueExtensions.kt
```diff
@@ -47,29 +52,39 @@ fun TextFieldValue.replaceSymbolText(
     newText: String,
 ): TextFieldValue {
     if (!text.contains(symbol)) return this
-    val firstSpaceIndexAfterCursor = text.indexOf(' ', startIndex = selection.end)
+    val firstSpaceOrNewLineIndexAfterCursor = getFirstSpaceOrNewLineIndexAfterCursor(
```
devotaaabel, line 55: Can you add kdoc to this function, and others in this file that might be a little nebulous? It's hard to tell what this does. It seems to be replacing the text after the symbol, but with what, it's hard to parse. Is it updating the text style? Maybe there's a clearer way to name this function?

JohnOberhauser: updated.  Let me know if it's clear enough

devotaaabel: Much better

devotaaabel (approved):

PR #423, opened 2024-01-28
Fixing button heights

JohnOberhauser:
- Fixing button heights throughout the app
- Fixing some button text sizes
- Setting default button vertical padding to 0 because it was cutting off text inside the buttons (the vertical padding doesn't affect the height of the button)


core/ui/common/src/main/java/org/mozilla/social/core/ui/common/button/MoSoButton.kt
```diff
@@ -190,10 +190,16 @@ interface MoSoButtonTheme {
...
+    val default = 48.dp
+    val sm = 40.dp
+    val xs = 34.dp
+}
+
 object MoSoButtonDefaults {
 
     private val ButtonHorizontalPadding = 24.dp
-    private val ButtonVerticalPadding = 14.dp
+    private val ButtonVerticalPadding = 0.dp
```
devotaaabel, line 202: This value was from the design system, and I think it might be more important for consistency and support of different text sizes to have hard-coded padding vs. a hardcoded height... Let's chat offline about this

devotaaabel (approved):

Issue #419, opened 2024-01-28
Visual issues on the profile tab


cadeyrn:
_I am aware that the app is still early in development, and some things are probably already on your radar anyway. To make sure nothing gets missed, I'll open a series of issues with things I've noticed._

There are a few visual issues in the profile tab, which are basically all minor things, but I get stuck with my eye every time.

1. The tab label "Posts and Replies" is on two lines.
2) I'm not entirely happy with how the statistics are displayed (Followers - Following - Posts). I'm not sure what the perfect implementation looks like, but it all seems rather pushed to the left. Maybe make the font smaller and put everything in one line like the official Mastodon app? But the official Mastodon app only shows two stats as it shows the number of posts on a different place, so that could be difficult in terms of space.
3. the "Edit Profile" button is quite big and sticks to the cover picture.

![Screenshot_2024-01-28-10-13-41-42_45d8e3eaae3374bfcb41fa9e7b6b444c](https://github.com/MozillaSocial/mozilla-social-android/assets/1100614/cea270f5-9090-42ae-b683-d34abf056f78)

JohnOberhauser:
Jira tickets SOCAND-431 and SOCAND-423

JohnOberhauser:
fixed in https://github.com/MozillaSocial/mozilla-social-android/pull/426

PR #426, opened 2024-01-28
Fixing tab issue

JohnOberhauser:
Tabs on smaller devices might have text that expands to two lines, like here https://github.com/MozillaSocial/mozilla-social-android/issues/419

This could also happen when we add in localization.

One possible solution would be to use the material3 `ScrollableTabRow`.  However, this looks bad when the tabs don't take up the entire space, because they are all squished in the middle or to the side.

My solution here is to copy `ScrollableTabRow` to our code and modify it so that minimum tab widths are dynamically calculated.  This makes it so that if your screen is big enough, the tabs look exactly as they do already, but if your screen is too small, you can scroll the tabs.

### Before

1706482215_grim

### After

1706482174_grim

### DIFF
--- a/core/ui/common/src/main/java/org/mozilla/social/core/ui/common/MoSoTab.kt
+++ b/core/ui/common/src/main/java/org/mozilla/social/core/ui/common/MoSoTab.kt
@@ -1,20 +1,52 @@
 package org.mozilla.social.core.ui.common
 
+import androidx.compose.animation.core.AnimationSpec
+import androidx.compose.animation.core.FastOutSlowInEasing
+import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.ScrollState
+import androidx.compose.foundation.background
+import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.selection.selectableGroup
+import androidx.compose.material3.ScrollableTabRow
+import androidx.compose.material3.Surface
 import androidx.compose.material3.Tab
-import androidx.compose.material3.TabPosition
 import androidx.compose.material3.TabRow
-import androidx.compose.material3.TabRowDefaults
-import androidx.compose.material3.TabRowDefaults.tabIndicatorOffset
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
+import androidx.compose.ui.draw.clipToBounds
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.SubcomposeLayout
+import androidx.compose.ui.platform.debugInspectorInfo
 import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.unit.Density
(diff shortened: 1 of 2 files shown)


core/ui/common/src/main/java/org/mozilla/social/core/ui/common/MoSoTab.kt
```diff
@@ -35,17 +67,278 @@ fun MoSoTabRow(
...
+                        tabPositions = tabPositions,
+                        selectedTab = selectedTabIndex
+                    )
+                }
+            }
+        }
+    }
+}
+
+private fun calculateTabMinWidth(
```
JohnOberhauser, line 161: I added this function to calculate the min width

core/ui/common/src/main/java/org/mozilla/social/core/ui/common/MoSoTab.kt
```diff
@@ -35,17 +67,278 @@ fun MoSoTabRow(
...
+        val scrollState = rememberScrollState()
+        val coroutineScope = rememberCoroutineScope()
+        val scrollableTabData = remember(scrollState, coroutineScope) {
+            ScrollableTabData(
+                scrollState = scrollState,
+                coroutineScope = coroutineScope
+            )
+        }
+
+        BoxWithConstraints {
```
JohnOberhauser, line 84: I wrapped the subcompose layout in a `BoxWithConstraints` so we can get the max width of the outer surface

core/ui/common/src/main/java/org/mozilla/social/core/ui/common/MoSoTab.kt
```diff
@@ -35,17 +67,278 @@ fun MoSoTabRow(
...
+                Modifier
+                    .fillMaxWidth()
+                    .wrapContentSize(align = Alignment.CenterStart)
+                    .horizontalScroll(scrollState)
+                    .selectableGroup()
+                    .clipToBounds()
+            ) { constraints ->
+                val tabMeasurables = subcompose(TabSlots.Tabs, tabs)
+
+                val minTabWidth = calculateTabMinWidth(
```
JohnOberhauser, line 95: Here is where we call the function to calculate the tab min width

devotaaabel (commented):
Have you chatted with design about this? I see in the kdoc for `ScrollableTabRow`, the following param and description: `edgePadding - the padding between the starting and ending edge of the scrollable tab row, and the tabs inside the row. This padding helps inform the user that this tab row can be scrolled, unlike a TabRow.` I think it's a valid concern that it might not be obvious that the tabs are scrollable without padding at the start/end, so I wonder if design would be ok with the padding at start and end if we just used `ScrollableTabRow` so we don't have to maintain this. That being said, this is a pretty cool customization if we want to go this route

JohnOberhauser:
> Have you chatted with design about this? I see in the kdoc for `ScrollableTabRow`, the following param and description: `edgePadding - the padding between the starting and ending edge of the scrollable tab row, and the tabs inside the row. This padding helps inform the user that this tab row can be scrolled, unlike a TabRow.` I think it's a valid concern that it might not be obvious that the tabs are scrollable without padding at the start/end, so I wonder if design would be ok with the padding at start and end if we just used `ScrollableTabRow` so we don't have to maintain this. That being said, this is a pretty cool customization if we want to go this route

You can see that padding in action here:
1706543468_grim
It has the default 50dp padding.  In this case, the tabs could fit the whole width easily, but the are squished together.  For this reason, I think my solution is the way to go

devotaaabel:
I'm worried about it being a UX issue if people can't tell the tabs are scrollable. I'll review this as-is but I think we need to talk to design about it at some point

core/ui/common/src/main/java/org/mozilla/social/core/ui/common/MoSoTab.kt
```diff
@@ -35,17 +67,279 @@ fun MoSoTabRow(
     },
     tabs: @Composable () -> Unit,
 ) {
-    TabRow(
-        selectedTabIndex = selectedTabIndex,
+    Surface(
```
devotaaabel, line 70: Can you please pull everything that was pulled from `ScrollableTabRow` into it's own file called `MoSoScrollableTabRow` (with AOSP license), and add comments that it's based on `ScrollableTabRow` but with the added functionality? So ideally all that will have to change in this file is updating `TabRow` to `MoSoScrollableTabRow`

JohnOberhauser: I moved it to it's own file and added the license.  I'm just going to keep it called `MoSoTabRow` though.  Otherwise we would have `MoSoTabRow` and `MoSoScrollableTabRow` that both do the exact same thing.

JohnOberhauser: Also, put them both in a new package

devotaaabel (approved):
LGTM if design is cool with it

PR #427, opened 2024-01-29
Edit post part 2

timc-mozilla:
Edit post working for text-based messages

next step is media and polls


timc-mozilla:
> I just want to confirm, this isn't enabled in the UI yet in this PR?

correct. ill double check, but the code to show it is in PostCard and should be commented out.

JohnOberhauser (approved):

PR #432, opened 2024-01-30
Search double tap

JohnOberhauser:
- Making it so if you tap the search tab icon while it's already selected, you will be sent to the search screen
- Increasing the height of the search bar.  The touch target was too small
- Fixing an issue with search bar alignment between the discover screen and the search screen

### DIFF
--- a/app/src/main/java/org/mozilla/social/ui/bottombar/MoSoNavigationBar.kt
+++ b/app/src/main/java/org/mozilla/social/ui/bottombar/MoSoNavigationBar.kt
@@ -184,6 +184,7 @@ interface BottomBarTab {
 
     val tabText: StringFactory
     val navigationDestination: Destination
+    // Destination for when the user taps the icon a second time
     val doubleTapDestination: Destination?
 }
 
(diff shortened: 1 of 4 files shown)


devotaaabel (changes requested):
I think you want to use [gestures](https://developer.android.com/jetpack/compose/touch-input/pointer-input/tap-and-press) for this

JohnOberhauser:
> I think you want to use [gestures](https://developer.android.com/jetpack/compose/touch-input/pointer-input/tap-and-press) for this

Moving conversation to slack thread

app/src/main/java/org/mozilla/social/ui/bottombar/MoSoNavigationBar.kt
```diff
@@ -178,6 +184,7 @@ interface BottomBarTab {
 
     val tabText: StringFactory
     val navigationDestination: Destination
+    val doubleTapDestination: Destination?
```
devotaaabel, line 187: Can you add comments here for clarity?

devotaaabel (approved):

PR #435, opened 2024-02-01
Media screen part 1

JohnOberhauser:
- Setting up the media screen
  - There is a still a bunch of missing stuff on the media screen, this is just setup


core/navigation/src/main/java/org/mozilla/social/core/navigation/NavigationDestination.kt
```diff
@@ -112,6 +116,44 @@ sealed class NavigationDestination(
...
 
+    data class Media(
+        val attachments: List<Attachment>,
+        val startIndex: Int = 0,
+    ) : NavigationDestination(
+        route = ROUTE,
+    ) {
+        fun NavController.navigateToMedia(navOptions: NavOptions? = null) {
+            val mediaBundle = MediaBundle(
+                // remove blur hashes because the characters in the hash can mess up serialization
```
devotaaabel, line 127: I'm not sure what this means. Are we sure we don't want to use blur hashes?

devotaaabel (approved):

PR #438, opened 2024-02-01
Zoom and pan images

JohnOberhauser:
- Adding the ability to zoom and pan images in the media screen

### DIFF
--- a/feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
+++ b/feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
@@ -1,19 +1,37 @@
 package org.mozilla.social.feature.media
 
+import androidx.compose.foundation.background
+import androidx.compose.foundation.gestures.detectTransformGestures
+import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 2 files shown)


feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
```diff
@@ -24,14 +33,64 @@ internal fun MediaScreen(
             .fillMaxSize()
             .systemBarsPadding()
     ) {
-        val attachment = mediaBundle.attachments.first()
+        val attachment = mediaBundle.attachments[mediaBundle.startIndex]
+
+        Column {
```
devotaaabel, line 38: I don't think this column needs to be here

devotaaabel (approved):
Nice, I was worried that was gonna be way harder

feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
JohnOberhauser: I'm gonna use it in the next PR

PR #448, opened 2024-02-06
Media screen, swipe between images

JohnOberhauser:
- Enabling swiping between images on the media screen

### DIFF
--- a/feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
+++ b/feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
@@ -200,10 +200,6 @@ private fun ZoomableImage(
                             .coerceIn(-translationLimitX..translationLimitX)
                         translationY = (translationY + panY - centroidTranslationY)
                             .coerceIn(-translationLimitY..translationLimitY)
-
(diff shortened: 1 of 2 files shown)


timc-mozilla (approved):

feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
```diff
@@ -108,17 +118,43 @@ private fun MediaScreen(
...
 
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+private fun ImagePager(
+    attachments: List<Attachment>,
+    pagerState: PagerState,
+) {
+    val scale = remember { mutableFloatStateOf(1f) }
+
+    HorizontalPager(
```
devotaaabel, line 129: Should there be different implementations for landscape/compact height class devices?

devotaaabel (approved):

feature/media/src/main/java/org/mozilla/social/feature/media/MediaScreen.kt
JohnOberhauser: hmm, maybe.  I'll look into that in my following PRs

PR #455, opened 2024-02-09
Bug 1879155 - Add Glean probe-scraper GitHub workflow

akkomar:
This is needed for new metrics to be published.

Requires https://github.com/mozilla/probe-scraper/pull/685 to land first.


devotaaabel:
@akkomar thanks for opening this, I think this is failing on CI because you don't have the secret keys setup in your fork... I see that the required PR is merged, so I'll reopen an identical PR. Thanks!

PR #457, opened 2024-02-12
Enables edit post feature

devotaaabel:
- enable edit post
- update cursor to be at end of text

@JohnOberhauser can you confirm that there's nothing else left to do for this? I feel like I'm missing something


JohnOberhauser:
I noticed a bug.  If you have a post with a mention, and then edit that post, the html for the mention gets stripped from the post.  I.e, the handle is no longer highlighted and you can't click it

